PR #14, opened 2015-01-16
Add ability to change columnLength

Willyham:
@sdepold

### DIFF
--- a/lib/storages/sequelize.js
+++ b/lib/storages/sequelize.js
@@ -33,6 +33,8 @@ var Sequelize = require(resolve('sequelize', { basedir: process.cwd() }));
  * to create if "model" option is not supplied
  * @param {String} [options.storageOptions.columnName='name'] name of table column
  * holding migration name
+ * @param {Number} [options.storageOptions.columnLength=255] character length of the column.
+ * For utf8mb4 charsets under InnoDB, you may need to set this <= 190
  */
 module.exports = redefine.Class({
   constructor: function (options) {
@@ -41,7 +43,9 @@ module.exports = redefine.Class({
       // note 'sequelize' or 'model' is required
       modelName: 'SequelizeMeta',
       // note 'tableName' (optional) also supported
-      columnName: 'name'
+      columnName: 'name',
+      // You may need to override this for utf8mb4 charsets
+      columnLength: 255
     }, this.options.storageOptions || {});
 
     if (!this.options.storageOptions.model && !this.options.storageOptions.sequelize) {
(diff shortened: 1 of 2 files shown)


Willyham:
I assume changes to umzug will not be backported into sequelize core, right?

sdepold:
So basically we would like to be able to configure the datatype of the column right?

Willyham:
It doesn't really couple this to Sequelize any more than it currently is. It's not unreasonable to assume that each different storage mechanism will require different options.

That being said, @sdepold is right, the underlying issue is one of configurable datatype, so maybe a better solution is to change that rather than explicitly changing the key length. I'm not sure how you'd like to add that level of abstraction.

sdepold:
Oh. Wow. I totally missed that this is a PR

sdepold:
Maybe it would be the better idea to actually pass the type of the column rather than the length of the varchar?

lib/storages/sequelize.js
```diff
@@ -41,7 +43,9 @@ module.exports = redefine.Class({
       // note 'sequelize' or 'model' is required
       modelName: 'SequelizeMeta',
       // note 'tableName' (optional) also supported
-      columnName: 'name'
+      columnName: 'name',
+      // You may need to override this for utf8mb4 charsets
+      columnLength: 255
```
sdepold: So instead of `columnLength` we could use `columnType` which can then either be `Sequelize.STRING(something)` or just a plain string containg the respective information. Opinion?

sdepold:
Btw. after this change we have to actually pass the options to umzug here: https://github.com/sequelize/cli/blob/master/lib/tasks/db.js#L224

sdepold:
Hi @Willyham,

in case you are providing a PR for the CLI either, would you consider this ticket as well ? https://github.com/sequelize/cli/issues/20

Willyham:
Sorry, just got around to this. I'll hopefully follow up with a CLI diff next week and then I'll take a look at that one too.

lib/storages/sequelize.js
```diff
@@ -41,7 +43,9 @@ module.exports = redefine.Class({
       // note 'sequelize' or 'model' is required
       modelName: 'SequelizeMeta',
       // note 'tableName' (optional) also supported
-      columnName: 'name'
+      columnName: 'name',
+      // You may need to override this for utf8mb4 charsets
+      columnType: new Sequelize.STRING(255)
```
sdepold: Can we just set this to the previous value which was `Sequelize.STRING` ?

sdepold:
Thanks for your excellent work :)

sdepold:
Released as 1.3.0

PR #16, opened 2015-02-05
added ActiveRecord like logging messages for each executed migration

alubbe:
see https://github.com/sequelize/cli/issues/92

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,6 +43,8 @@ var Umzug = module.exports = redefine.Class({
       })
       .then(function (options) {
         return Bluebird.each(options.migrations, function (migration) {
+          var name = path.basename(migration.file, path.extname(migration.file));
+          var startTime;
           return self
             ._wasExecuted(migration)
             .catch(function () {
@@ -60,6 +62,14 @@ var Umzug = module.exports = redefine.Class({
                   params = params();
                 }
 
+                if (options.method === 'up') {
+                  console.log("== " + name + ": migrating =======");
+                } else {
+                  console.log("== " + name + ": reverting =======");
+                }
+
+                startTime = new Date();
+
                 return fun.apply(migration, params);
               }
             })
@@ -68,6 +78,14 @@ var Umzug = module.exports = redefine.Class({
                 return Bluebird.resolve(self.storage.logMigration(migration.file));
               } else if (options.method === 'down') {
                 return Bluebird.resolve(self.storage.unlogMigration(migration.file));
+              }
+            })
+            .tap(function () {
+              var duration = ((new Date() - startTime) / 1000).toFixed(3);
+              if (options.method === 'up') {
+                console.log("== " + name + ": migrated (" + duration +  "s)\n");
+              } else {
+                console.log("== " + name + ": reverted (" + duration +  "s)\n");
               }
             });
         });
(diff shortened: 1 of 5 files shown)


rpaterson:
Please don't output directly to console.log - use some logging system so that clients can control where the log goes.  E.g. Sequelize accepts an option { logging: <function> } and uses that to output logs.

alubbe:
does umzug already have a ```logging:``` option?
afaik, the sequelize-cli disables the sequelize logs (mainly to disable the sql prints), so I don't think I can use their logging function.

What would you propose?

rpaterson:
I'm not suggesting using the sequelize logs directly.  What I meant was add a `logging:` option to Umzug if it doesn't have one already.  Sequelize/cli can specify `logging: console.log` and other clients of Umzug can do as they see fit.

alubbe:
I like that idea - see newest commit

index.js
```diff
@@ -184,6 +207,12 @@ var Umzug = module.exports = redefine.Class({
     }
   },
 
+  log: function() {
+    if (this.options.logging) {
+      this.options.logging.apply(null, [].slice.call(arguments));
```
rpaterson: What do you expect to happen when the logging method is called with multiple arguments?

rpaterson:
Looks good to me, just missing documentation.

sdepold:
Will review it later today

index.js
alubbe: I'm not sure I fully understand your question. The consumer could potentially add their very own logging function that takes multiple arguments, or even want to pass multiple arguments into console.log. In both cases I would expect that .log passes them on 1-1 to the passed logging function. Does that answer your question?

alubbe:
@rpaterson does the documentation just go into the readme or somewhere else, as well? Do you want me to write something up?

index.js
rpaterson: A user of Umzug who is writing a logging method needs to know what to do with multiple arguments - e.g. concat them all with .toString?  Or is the first argument a context object, the second a template, and subsequent args params as per Bunyan logging?  Conversely the Umzug code needs to know when and how to call the log method with multiple args.  I think probably easiest would be just to have one argument that is a string - that's all you're using today AFAICT anyway.

rpaterson:
Just the readme AFAIK

index.js
alubbe: understood and agreed. let's simplify for now and add more arguments, such as a possible options arguments, only when actually needed

sdepold:
Looks good. The readme is right. Please also add tests for the log function :)

alubbe:
doc & tests added

rpaterson:
LGTM

sdepold:
Sorry to bother you but would it be possible to actually call the migrate function and check if the expected logging behavior is executed?

alubbe:
Okay here we go

alubbe:
@sdepold feedback?

sdepold:
Sorry. The last days were a bit stressful :) Could you please add an assertion for the actual arguments that get passed to logger? So ideally we would really make messages are getting logged.

sdepold:
Hossa! Merged + added the missing test pieces on my own.

sdepold:
just released v1.5.0

alubbe:
awesome!

PR #21, opened 2015-02-19
add a Sequelize option

### LINKED ISSUE #20: Reference to Sequelize
It's possible pass the reference to Sequelize in the sequelize storage constructor instead of doing this?

[https://github.com/sequelize/umzug/blob/master/lib/storages/sequelize.js#L8](https://github.com/sequelize/umzug/blob/master/lib/storages/sequelize.js#L8)
```js
var Sequelize = require(resolve('sequelize', { basedir: process.cwd() }));
```

I'm working on a framework that depends on this library, but the applications using the framework doesn't depends directly on sequelize.

[https://github.com/axelhzf/barbakoa/blob/master/app/server/migrations.js](https://github.com/axelhzf/barbakoa/blob/master/app/server/migrations.js)

Does this make sense?

axelhzf:
fixes #20

### DIFF
--- a/lib/storages/sequelize.js
+++ b/lib/storages/sequelize.js
@@ -4,8 +4,7 @@ var _         = require('lodash');
 var Bluebird  = require('bluebird');
 var path      = require('path');
 var redefine  = require('redefine');
-var resolve   = require('resolve').sync;
-var Sequelize = require(resolve('sequelize', { basedir: process.cwd() }));
+var resolve   = require('resolve');
 
 /**
  * Sequelize storage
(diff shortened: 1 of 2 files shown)


lib/storages/sequelize.js
```diff
@@ -43,11 +42,23 @@ module.exports = redefine.Class({
...
     }, this.options.storageOptions || {});
-
+    
+    if(!this.options.storageOptions.columnType) {
+      // You may need to override this for utf8mb4 charsets
+      var Sequelize;
+      try {
+        Sequelize = require(resolve.sync('sequelize', { basedir: process.cwd() }))
+      } catch(e) {
+        Sequelize = this.options.storageOptions.Sequelize;
```
sdepold: I wonder if we should check for this option first?

sdepold:
One thing that came into my mind tonight: Couldn't we just `sequelize.constructor` ?

axelhzf:
Yes, to use `sequelize.constructor` is a cleaner solution. I have modified the PR

lib/storages/sequelize.js
```diff
@@ -4,8 +4,7 @@ var _         = require('lodash');
 var Bluebird  = require('bluebird');
 var path      = require('path');
 var redefine  = require('redefine');
-var resolve   = require('resolve').sync;
-var Sequelize = require(resolve('sequelize', { basedir: process.cwd() }));
+var resolve   = require('resolve');
```
sdepold: I think we don't need this anymore?

lib/storages/sequelize.js
```diff
@@ -56,6 +53,7 @@ module.exports = redefine.Class({
     if (!this.options.storageOptions.model) {
       var sequelize = this.options.storageOptions.sequelize;
       var modelName = this.options.storageOptions.modelName;
+      var columnType = this.options.storageOptions.columnType || sequelize.constructor.STRING;
```
sdepold: Could you do me a favor and add a line `var Sequelize = sequelize.constructor` ?

lib/storages/sequelize.js
axelhzf: yes, you're right

lib/storages/sequelize.js
axelhzf: Yes, I just push a new commit with the two changes

sdepold:
Awesome :)

PR #28, opened 2015-06-02
Update sequelize storage to sort results.

PatriotBob:
When the number of migrations start to increase and SequelizeMeta is persisted in a proper db, in my case postgresql, the table can start to be broken up across pages. This causes the results of findAll to be potentially out of order.
I'm using [Sequelize-cli](https://github.com/sequelize/cli) to build my migrations so all my scripts are prefixed with a timestamp making this default perfect, but if this doesn't work in all cases it would be extend this to use a storage option and update [Sequelize-cli](https://github.com/sequelize/cli) to use this sort by default.

### DIFF
--- a/lib/storages/sequelize.js
+++ b/lib/storages/sequelize.js
@@ -119,7 +119,7 @@ module.exports = redefine.Class({
     return this._model()
       .sync()
       .then(function(Model) {
-        return Model.findAll();
+        return Model.findAll({ order: [ [ self.options.storageOptions.columnName, 'ASC' ] ] });
       })
       .then(function(migrations) {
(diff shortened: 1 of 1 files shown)


rpaterson:
LGTM

sdepold:
Makes sense. Should there be a default ?

srussellextensis:
We ran into an issue with the downgrade migrations being run in a different order than the migrations were listed in the SequelizeMeta table, and it caused issues with our schema migrations.

I think this change would fix that problem, but for now we're going to sort the list of migrations in our code and explicitly specify the downgrade order.

sdepold:
ah there is a default value already :dooh:

PR #40, opened 2015-11-27
Added restrictions to `up|down({ to: ... })`; fixes #36

### LINKED ISSUE #36: umzug.up({to: '...'}) runs all pending migrations instead of only those from the current state to the given name
According to the readme:

> It is also possible to pass the name of a migration in order to just run the migrations from the current state to the passed migration name.
>
>     umzug.up({ to: '20141101203500-task' }).then(function (migrations) {});

However, this doesn't appear to work as described. Using umzug 1.6.0, when I pass an options object like `{to: '201510301204-add-index'}`, umzug runs _all_ pending migrations, not just the migrations between the current state and `201510301204-add-index`. It also runs all pending migrations if I pass a completely bogus migration name like `{to: 'bogus'}`.

I peeked at the unit tests to see if they held any clues, but passing an options object to `up()` doesn't seem to be tested. Does this feature actually exist?

jukkah:
- Check that there is a migration that matches to `to` parameter.
- Check that the migration is pending/executed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,6 +155,25 @@ var Umzug = module.exports = redefine.Class({
     } else {
       return this.pending().bind(this)
         .then(function (migrations) {
+          var result = Bluebird.resolve().bind(this);
+
+          if (options.to) {
+            result = result
+              .then(function () {
+                // There must be a migration matching to options.to...
+                return this._findMigration(options.to);
+              })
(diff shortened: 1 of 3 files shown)


sdepold:
nice

sdepold:
v1.7.1

Issue #39, opened 2015-11-26
Squashed migrations


jukkah:
Is there any way to combine migrations to squashed migrations and run them instead of the original migrations? Currently no, I guess.

As you know, if there is a zillion of migrations, running them all from the beginning to the end may take a while. In addition, if you have software@1.2 and you upgrade it to software@2.0, performance of the database migration matters.

For example, [Django](https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations) have some kind of implementation for squashed migrations. I'm not sure if Django replaces migrations with squashed ones, but I think it's better to save them just in case. So there should be all the migrations **plus** some squashed migrations.

![migrations](https://cloud.githubusercontent.com/assets/1504091/11429449/88bc5a8c-9480-11e5-929f-ba4888fcc3a1.png)

In the picture above, black dots are migrations created between releases. Bellow the migrations are squashed migrations. In this example, there are squashes for each minor and major release and one from the beginning to the latest release.

I think, in umzug, we need only
- a list of migrations combined into each squashed migration,
- a way to figure out what squash to use if any, and
- a way to make difference between migrations and squashes.

I can implement this feature but it needs some discussion first.

sdepold:
How would this mechanism be applied? Is that a runtime feature that is executed when running `up` or `down` or is it an extra command that squashes the existing migrations?

jukkah:
This would be a runtime feature. Because we don't care about implementations of `up` and `down` functions, we don't need to analyze them to figure out how to squash them. If some other libraries can create migrations they should also squash them.

My vision of squash file is to follow syntax of migration file:
```javascript
'use strict';

var Bluebird = require('bluebird');

module.exports = {
  // List of migrations squashed in this squash file.
  migrations: ['some-id', 'some-other-id'],

  up: function () {
    return new Bluebird(function (resolve, reject) {
      // Describe how to achieve the task.
      // Call resolve/reject at some point.
    });
  },

  down: function () {
    return new Bluebird(function (resolve, reject) {
      // Describe how to revert the task.
      // Call resolve/reject at some point.
    });
  }
};
```

When running `up` and `down` we (un)log each of listed migrations.

These files should be stored in `squashes` sub-directory or something like that. Another way is to use file names not starting with number. This allows us to require them all at the beginning and then figure out the "best way".

sdepold:
tbh I don't see a major benefit unless we are also implementing this:

```
[...]

PR #41, opened 2015-12-08
Squashed migrations; #39

jukkah:
Sorry for too large commit. :smile: There is still one open question:

How should we handle squashes with interrupted queue of migrations? That's we have a squash for migrations `20151208080000.js` and `20151208120000.js` but later someone publishes a migration `20151208100000.js`. Currently, the squash cannot be applied at all because it was not clear to me if the squash should be run before that migration or vice versa.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,13 @@ var Umzug = module.exports = redefine.Class({
       wrap:    function (fun) { return fun; }
     }, this.options.migrations);
 
+    this.options.squashes = _.assign({
+      params:  [],
+      path:    path.resolve(process.cwd(), 'migrations/squashes'),
+      pattern: /^[\w-]+\.js$/,
+      wrap:    function (fun) { return fun; }
+    }, this.options.squashes);
+
     this.storage = this._initStorage();
   },
 
@@ -44,11 +51,13 @@ var Umzug = module.exports = redefine.Class({
         return self._findMigration(migration);
       })
       .then(function (migrations) {
+        return self._applySquashes(migrations);
+      })
+      .then(function (migrations) {
         return _.assign(options, { migrations: migrations });
       })
       .then(function (options) {
         return Bluebird.each(options.migrations, function (migration) {
-          var name = path.basename(migration.file, path.extname(migration.file));
           var startTime;
           return self
             ._wasExecuted(migration)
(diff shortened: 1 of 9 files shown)


sdepold:
omg so much code

sdepold:
I would totally appreciate a review from someone else. I'm still not entirely sure what to think about this functionality. @mickhansen @janmeier @durango @overlookmotel

sdepold:
Pinging @mickhansen @janmeier @durango @overlookmotel one more time :)

mickhansen:
I don't use the migrations so i don't really have an opinion here.
What's the usecase for squashing migrations?

overlookmotel:
Sorry I've been slow to reply on this. Is there an issue which describes what this PR is trying to achieve? There's quite a lot of code to wade through and I'm not clear on what it's trying to do!

I haven't used migrations myself. I've wanted to, but every time I've looked at it, my thought has always been that surely there's a better way of doing this (like an intelligent diff between current database state and the models defined). So I'd be pleased to see anything that makes migrations a bit easier to use.

jukkah:
@overlookmotel #39. As described in that issue [Django](https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations) has a great implementation of squashed migrations. In addition, there are many no-Node.js enterprise orms/frameworks supporting this.

Lets say, we are developing a massive ERP system. Some of our customers are still using v1.0 while we are publishing v10.0. If they decide to update their self-hosted system, there will be a huge diff on DB. It might be impossible to update the DB by the intelligent diff without loosing data. If we use regular migrations, it will take a year to run all the zillion of migrations. But if we have squashed these migrations already, they need to run only 9 squashes (= optimized versions of all migrations between major versions) (v1.0→v2.0→v3.0→v4.0→v5.0→v6.0→v7.0→v8.0→v9.0→v10.0). In case of you cannot understand the speed of squashes, figure out the speed of applying MySQL dump file. It is a real squash (in SQL level) from empty DB to the point when it was taken.

jukkah:
@sdepold It seems the other guys are not familiar enough with migrations to give their ±1 here. In addition, this is a big feature maybe too big to merge straight to master this way. Should we close this PR for now meanwhile I continue development in my fork until I consider it stable enough to be merged?

sdepold:
@jukkah Guess that's a good idea :) Feel free to re-open this PR when you feel like it :)

PR #54, opened 2016-02-02
give more details on failing to load custom storage

timruffles:
It's annoying when it's failing to load due to syntax error, and you assume it's path (especially as path is only info given).


sdepold:
Do you think you could add a test for it?

sdepold:
released as 1.8.1

PR #60, opened 2016-02-12
Extracted common parts of `up` and `down` to new `_run`

jukkah:
Refactored `up` and `down` methods since they have had a lot of common code. Diff is not showing changes in the best way, so I'll explain it shortly. There is now `_run(bool: up, any: options, fn: rest)` and `up` is calling it as `this._run(true, options, this.pending.bind(this))` while `down` calls it as `this._run(false, options, getExecuted.bind(this))`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,56 +130,7 @@ var Umzug = module.exports = redefine.Class({
   },
 
   up: function (options) {
-    if (typeof options === 'string') {
-      return this.up([ options ]);
-    } else if (Array.isArray(options)) {
-      return Bluebird.resolve(options).bind(this)
-        .map(function (migration) {
-          return this._findMigration(migration)
-        })
-        .then(function (migrations) {
-          return this._arePending(migrations);
-        })
-        .then(function () {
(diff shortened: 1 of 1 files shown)


jukkah:
@sdepold All tests passing.

index.js
```diff
@@ -191,20 +142,30 @@ var Umzug = module.exports = redefine.Class({
...
+          : Bluebird.resolve([]);
       });
-    } else if (typeof options === 'string') {
-      return this.down([ options ]);
+    } else {
+      return this._run(false, options, getExecuted.bind(this));
+    }
+  },
+
+  _run: function(up, options, rest) {
```
sdepold: Can we call the first arg `method` or `direction`. I dislike boolean arguments quite a lot :)

sdepold:
rest lgtm :)

index.js
jukkah: Yep, I'll change it.

jukkah: Done.

sdepold:
:cool:

PR #61, opened 2016-02-12
Refactor tests

jukkah:
Grouped tests under the same describes. See changes to output in Travis.

Note that most of changes are space changes (decreased indentation). I isolated them to the second commit while the first commit is only one worth to review.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1,0 +1,4 @@
+'use strict';
+
+require('./index');
+require('./storages');
--- a/test/test.js
+++ b/test/test.js
@@ -1,4 +1,0 @@
-'use strict';
-
-require('./index');
-require('./storages');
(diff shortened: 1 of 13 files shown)


sdepold:
There is one failing test now

jukkah:
@sdepold All tests passing.

test/test.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
```
sdepold: What is this file for?

jukkah: It is main test file but it is not required.

sdepold: Since we have the gulp task it seems redundant to me :) Can we remove it?

jukkah: Done.

PR #63, opened 2016-03-11
Fixed #50; `down({})` reverts only the latest migration

### LINKED ISSUE #50: revert last migration - reverts all if param is empty object
Since introduction of `to: 0` to revert all migrations I think a bug was introduced;

``` javascript
umzug.down({}).then(function (migration) {
   // will be same as using {to: 0} i.e. will revert ALL migrations
});
```

using `undefined` as params works as expected. If this is intentional, I think readme should be updated so users will know of this as well.

jukkah:
Fixing #50.

### DIFF
--- a/index.js
+++ b/index.js
@@ -140,7 +140,7 @@ var Umzug = module.exports = redefine.Class({
       });
     }.bind(this);
 
-    if (typeof options === 'undefined') {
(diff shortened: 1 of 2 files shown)


sdepold:
Cool :bow:

sdepold:
Released as 1.9.1 :)

PR #64, opened 2016-03-16
#50: Ignore unkown options in down()

jukkah:
Fixed problem found in 7cc0f2f.

### DIFF
--- a/test/index/down.test.js
+++ b/test/index/down.test.js
@@ -77,27 +77,6 @@ describe('down', function () {
 
     describe('when no option is specified', function () {
       beforeEach(function () {
-        return this.umzug.down().bind(this).then(function (migrations) {
-          this.migrations = migrations;
-        });
-      });
-
-      it('returns 1 item', function () {
-        expect(this.migrations).to.have.length(1);
-        expect(this.migrations[0].file).to.equal(this.migrationNames[2] + '.js');
-      });
-
-      it('removes the reverted migrations from the storage', function () {
-        return this.umzug.executed().bind(this).then(function (migrations) {
-          expect(migrations).to.have.length(2);
-          expect(migrations[0].file).to.equal(this.migrationNames[0] + '.js');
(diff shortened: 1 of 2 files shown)


tremby:
It seems rather overcomplicated. What about this?

```js
// Deal with special case where an array has been passed
if (Array.isArray(options)) {
  return this._run('down', options, getExecuted.bind(this));
}

if (typeof options === "undefined" || options === null || options.to == null) {
  return getExecuted().bind(this).then(function(migrations) {
    if (migrations[0]) {
      return this.down(migrations[0].file);
    }
    return Bluebird.resolve([]);
  });
}
```

Untested.

tremby:
No, ignore the above. It doesn't handle some things properly. It think this would be better handled in the _run method.

tremby:
I was thinking something more along the lines of this: https://github.com/tremby/umzug/commit/0db16eba7a754542504b570d4a226d306f0fcc58

It's untested (see #65); let me know what you think. It may well be incomplete.

tremby:
You can ignore that patch for now; it's entirely broken. Too tired to hack on it right now, though.

jukkah:
Even if it is broken, I got your point :sunglasses:

jukkah:
@sdepold Your review is needed.

test/index/down.test.js
```diff
@@ -77,27 +77,6 @@ describe('down', function () {
 
     describe('when no option is specified', function () {
       beforeEach(function () {
-        return this.umzug.down().bind(this).then(function (migrations) {
-          this.migrations = migrations;
-        });
-      });
-
-      it('returns 1 item', function () {
```
sdepold: why are we deleting tests?

jukkah: I have considered `down()` and `down({})` are the same test cases. In addition I'm refactoring this completely.

jukkah:
I give up with this as I'm working on v2.0 now.

PR #68, opened 2016-03-24
Fixed nesting of tests

jukkah:
Some test cases were required before corresponding `index.js` and therefore ran
outside their container. For example, _constructor_ were before beginning of
_Umzug_ in [test results][1]. See lines [389][2] and [471][3].

[1]: https://travis-ci.org/sequelize/umzug/jobs/117678313
[2]: https://travis-ci.org/sequelize/umzug/jobs/117678313#L389
[3]: https://travis-ci.org/sequelize/umzug/jobs/117678313#L471

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,7 +25,7 @@ gulp.task('lint', function () {
 
 gulp.task('test', function () {
   gulp
-    .src(path.resolve(__dirname, 'test', '**', '*.js'), { read: false })
+    .src(path.resolve(__dirname, 'test', '**', 'index.js'), { read: false })
     .pipe(mocha({
        reporter:    'spec',
        ignoreLeaks: true,


jukkah:
@sdepold Cosmetic change but I don't like merging anything without someone's :+1:.

sdepold:
> @sdepold Cosmetic change but I don't like merging anything without someone's .

thanks :)

sdepold:
@jukkah Isn't the problem that we need to change the test structures in the test itself? https://github.com/sequelize/umzug/blob/112c7258a8bbabf9863e06467bd7fad8499342bc/test/index/constructor.test.js#L7. The referenced log line is just not nested under Umzug because the respective file doesn't specify it.

jukkah:
>  Isn't the problem that we need to change the test structures in the test itself?

Yep, it is.

> The referenced log line is just not nested under Umzug because the respective file doesn't specify it.

We can't use `describe('Umzug', ...)` as a wrapper in every file since it produces multiple Umzugs instead of combining them all together under a single Umzug. (Tested some months ago.) In addition, I decided to not wrap all files with `module.exports = function () { ... }` as it makes impossible to run a single test file.

The solution is to describe one Umzug and then require all files within it. To work correctly, _index.js_ needs to be required first before other files in the same directory.

Expected (index.js is required first and rest in alphabetical order):
```
Umzug         index.js
- constructor constructor.js
- down        down.js
- up          up.js
```

Actual (all files are required in alphabetical order):
```
- constructor constructor.js
- down        down.js
Umzug         index.js
- up          up.js```

sdepold:
Oh now I got the point of this change. I thought it was about JSCS and JSHint. Sure. Sure. Sure. You can merge as it is (or do whatever changes you want :D)

Issue #67, opened 2016-03-21
Tests currently fail


tremby:
I have node v5.9.0 and npm 3.7.3. Tests currently fail for me on the master branch (with the missing jshint dependency added):

```
  106 passing (1s)
  2 failing

  1) storages sequelize constructor accepts a "sequelize" option and creates a model:
     Error: expected undefined to sort of equal null
    at Assertion.assert (node_modules/expect.js/index.js:96:13)
    at Assertion.eql (node_modules/expect.js/index.js:230:10)
    at test/storages/sequelize.test.js:56:52
    at bound (domain.js:287:14)
    at runBound (domain.js:300:12)
    at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (node_modules/bluebird/js/release/promise.js:503:31)
    at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:560:18)
    at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:605:10)
    at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:684:18)
    at Async._drainQueue (node_modules/bluebird/js/release/async.js:126:16)
    at Async._drainQueues (node_modules/bluebird/js/release/async.js:136:10)
    at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/release/async.js:16:14)
  

  2) storages sequelize constructor accepts a "columnType" option:
     Error: expected undefined to sort of equal null
    at Assertion.assert (node_modules/expect.js/index.js:96:13)
    at Assertion.eql (node_modules/expect.js/index.js:230:10)
    at test/storages/sequelize.test.js:120:52
    at bound (domain.js:287:14)
    at runBound (domain.js:300:12)
    at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (node_modules/bluebird/js/release/promise.js:503:31)
    at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:560:18)
    at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:605:10)
    at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:684:18)
    at Async._drainQueue (node_modules/bluebird/js/release/async.js:126:16)
    at Async._drainQueues (node_modules/bluebird/js/release/async.js:136:10)
[...]

PR #69, opened 2016-03-24
Fixed tests to accept changed made in sequelize@3.15.1; Fixing #67

jukkah:
Sequelize@3.15.1 changes default `defaultValue` from `null` to `undefined` at
least for sqlite.

Another possible soolution is to delete these expections.

Fixing #67


jukkah:
It still looks good enough to myself after a couple of weeks, so I consider this approved.

Issue #70, opened 2016-03-25
What is the meaning of 'upName' and 'downName'?


Maddoc42:
The README mentions the two config parameters `upName` and `downName`:

```javascript
// The name of the positive method in migrations.
upName: 'up',

// The name of the negative method in migrations.
downName: 'down',
```

Am I wrong to assume they can be used to configure the methods that will be called in the migration files (default being `up` and `down`)? It doesn't seem to work for me, nor can I find any usage of those params in the umzug code.

My config:
```javascript
this.umzug = new Umzug({
      storage: 'sequelize',
      storageOptions: {
        sequelize,
        modelName: 'SeedsMeta',
        tableName: 'SeedsMeta',
      },
      upName: 'plant',
      downName: 'burn',
      logging: msg => logger.info(msg),
      migrations: {
        params: [sequelize.getQueryInterface(), require('sequelize')],
        path: 'seeders',
        pattern: /^\d+-[\w-]+\.js$/,
      },
    });
```

This calls the `up` and `down` methods instead of the expected `plant` and `burn`.

Any hint?

jukkah:
It seems this is a clear bug. Do you have any plans to fix this?

Maddoc42:
Thanks for getting back. I don't have any immediate plans to fix this, as the workaround is pretty simple and acceptable for now. If time permits, would you accept a pull request regarding this?

jukkah:
Pull requests are always welcome. :smile: I'll wait for your PR but if you don't have enough time, please tell us that someone else (who has more time) can fix this.

jukkah:
@Maddoc42 Fixed in `v1.10.0`.

PR #71, opened 2016-03-30
Fix upName / downName options being ignored

Maddoc42:
This PR should fix the `upName` / `downName` options of `Umzug` being ignored. See #70.


jukkah:
LGTM. This should fix the issue correctly.

PR #72, opened 2016-04-12
Introduce a 'from' option for up() method?

thanpolas:
In my workflow, developing a desktop native app using Electron, I need the functionality were migrations are executed from a specific migration and onwards.

The case is simple, my app gets installed on new machines. Something new comes up for the next version and so I create `migration-1`.

When I will release the next version which includes the `migration-1` script, i want this migration to run on all clients that have the app already installed and not run on new installations of the new version which incorporate the migration part in the codebase / db schema already.

so would you accept a `from` option for the up() method?

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,11 @@ umzug.up({ from: '20141101203500-task' }).then(function (migrations) {});
 
 In the above example umzug will execute all the pending migrations found **after** the specified migration. This is particularly usefull if you are using migrations on your native desktop application and you don't need to run past migrations on new installs while they need to run on updated installations.
 
+You can combine `from` and `to` options to select a specific subset:
+
+```js
+umzug.up({ from: '20141101203500-task', to: '20151201103412-items' }).then(function (migrations) {});
+```
 
 Running specific migrations while ignoring the right order, can be done like this:
 
--- a/README.md
+++ b/README.md
@@ -193,6 +193,15 @@ It is also possible to pass the name of a migration in order to just run the mig
 umzug.up({ to: '20141101203500-task' }).then(function (migrations) {});
 ```
 
+You also have the ability to choose to run migrations *from* a specific migration, excluding it:
+
+```js
+umzug.up({ from: '20141101203500-task' }).then(function (migrations) {});
+```
+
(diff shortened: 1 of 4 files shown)


thanpolas:
@jukkah ready for review

jukkah:
Before reading the code carefully, have you considered what happens by calling

- `up({from: a, to: d})`
- `up({from: a})` when _a_ and _b_ are already executed but _c_, _d_, ... are pending
- `up({from: b})` when _a_, _b_, ... are all pending

thanpolas:
Yes, in the flow of determining which migrations to run, the 'FROM' runs first (if it exists) filtering migrations and then passes it on to the 'TO' option if it exists...

The FROM algo will only return pending or executed migrations depending on the method (up/down)

index.js
```diff
@@ -211,6 +219,47 @@ var Umzug = module.exports = redefine.Class({
...
+        var found = false;
+        return migrations.filter(function(migration) {
+          if (migration.testFileName(from)) {
+            found = true;
+            return false;
+          }
+          return found;
+        });
+      })
+      .then(function(migrations) {
```
jukkah: What is the point of this?

index.js
```diff
@@ -211,6 +219,47 @@ var Umzug = module.exports = redefine.Class({
...
+          })
+          .catch(function() {
+            if (method === 'up') {
+              return true;
+            } else {
+              return false;
+            }
+          });
+      })
+      .then(function(migrations) {
```
jukkah: The same here too.

README.md
```diff
@@ -193,6 +193,15 @@ It is also possible to pass the name of a migration in order to just run the mig
...
 
+You also have the ability to choose to run migrations *from* a specific migration, excluding it:
+
+```js
+umzug.up({ from: '20141101203500-task' }).then(function (migrations) {});
+```
+
+In the above example umzug will execute all the pending migrations found **after** the specified migration. This is particularly usefull if you are using migrations on your native desktop application and you don't need to run past migrations on new installs while they need to run on updated installations.
+
+
```
jukkah: Should we have an additional example like `umzug.up({ from: '...', to: '...' })`?

thanpolas:
applied fixes for your remarks

jukkah:
LGTM

PR #75, opened 2016-04-21
Add function level jsdoc documentation

jukkah:
Add some function level jsdoc documentation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,37 @@ var Migration = require('./lib/migration');
 var path      = require('path');
 var redefine  = require('redefine');
 
-var Umzug = module.exports = redefine.Class({
+/**
+ * @class Umzug
+ */
+var Umzug = module.exports = redefine.Class(/** @lends Umzug.prototype */ {
+  /**
+   * Constructs Umzug instance.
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -117,22 +165,44 @@ var Umzug = module.exports = redefine.Class({
 
         return all.filter(function (migration) {
           return executedFiles.indexOf(migration.file) === -1;
-        }).sort(function (a, b) {
```
jukkah: What was this sort doing? Nothing, I guess...

sdepold: it ensures the order of files?

jukkah: Yep. I can't figure out any cases where `this._findMigrations()` resolves to a list of migrations already not in order. Otherwise, the sort should be in `_findMigrations()`?

sdepold: Fair point.I wonder if I actually added that due to a bug report. Maybe windows problems?

jukkah: It should not be a bug fix since it has introduced and last changed in #1. But it could be for Windows. As it makes no harm, I'll move it to `_findMigrations`.

PR #76, opened 2016-04-29
Make Umzug an EventEmitter, add four events.

crrobinson14:
The attached PR adds EventEmitter functionality to Umzug. Unit tests and documentation for four useful events are included.

Rationale: My application has some logging and consistency-checking requirements for each migration that go beyond what can be conveniently done via the `logging` option. This leads to a lot of boilerplate code in migrations, to the point where almost 50% of a typical migration is this boilerplate. This makes migrations a bit ugly to read, and it's more lifting for new team members.

Using the event emitter pattern could let us eliminate this boilerplate by giving us more programmatic/event-driven insight into the individual migrations Umzug was running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ Using the [`sequelize` storage](lib/storages/sequelize.js) will create a table i
 }
 ```
 
+#### Events
+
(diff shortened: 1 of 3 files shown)


crrobinson14:
Strange... `npm test` works great locally...

jukkah:
LGTM :+1:

Issue #79, opened 2016-05-26
Is there a way to programmatically define migrations?


techwraith:
I'd like to generate migration up and down functions programmatically (without writing to the file system) is there a way to do that?

jukkah:
I'm not 100% sure what you mean. Do you mean
```javascript
let migration = {
  up(...) { ... },
  down(...) { ... }
};
umzug.execute({ migrations: [migration], method: 'up' });
```
instead of
```javascript
umzug.execute({ migrations: ['migration1.js'], method: 'up' });
```
? This is not currently possible but pull request is more than welcome.

hackhat:
Any news on this? 

My code is in typescript, therefore migrations are in typescript as well. Is there anyway to pass the migrations as an object instead of providing the folder for umzug to lookup?

jedwards1211:
@Techwraith as a result of https://github.com/sequelize/umzug/pull/164 it is possible to pass an array of `Migration` class instances in `options.migrations`.

hackhat:
Update: If anyone is interested in how I fixed... is I build my own migration runner in plain typescript.

wildcookie007:
> Update: If anyone is interested in how I fixed... is I build my own migration runner in plain typescript.

Would you please share your solution?

sushantdhiman:
https://github.com/sequelize/umzug/pull/199 introduces helper which can be used to build migrations array

PR #85, opened 2016-06-28
Support ES6 module defaults with Babel6

LegNeato:
In my setup, I had my migrations as the default export of an ES6 module
that was transpiled with Babel and it didn't work. This fixes it.

See http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default.

I wasn't sure how to test this without introducing babel, so I faked it a bit.
Let me know if you have better ideas.

Take this as another vote for https://github.com/sequelize/umzug/issues/79.
`umzug` really should leave module resolution to the end user...that way it
doesn't have to support everything under the sun (coffeescript, babel, typescript).

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -97,6 +97,9 @@ module.exports = redefine.Class(/** @lends Migration.prototype */{
    */
   _exec: function (method, args) {
     var migration  = this.migration();
+    if (migration.default) {
+      migration = migration.default;
+    }
     var fun        = migration[method] || Bluebird.resolve;
     var wrappedFun = this.options.migrations.wrap(fun);
 
--- a/test/index/execute.test.js
+++ b/test/index/execute.test.js
@@ -182,6 +182,40 @@ describe('coffee-script support', function () {
   });
 });
 
+describe('ES6 module support', function () {
+  beforeEach(function () {
+    helper.clearTmp();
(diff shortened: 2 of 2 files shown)


jukkah:
This looks like some kind of babel specific hack. -1 for current implementation.

I wonder if it would be possible to make configurable package resolver option. By default, it will do the same as [Migration.migration](https://github.com/sequelize/umzug/blob/master/lib/migration.js#L46). It should give programmers more power over their migrations.

jukkah:
As an example, one can write something like this:

```javascript
new Umzug({
  resolve(packageName) {
    // require(packageName) and so on
    // return exported module
  }
})
```

jamesreggio:
If you're looking to use Babel's ES6 module syntax right now, you can do this:

```javascript
export async function up() {
  console.log('We are up!');
  return Promise.delay(2500);
}

export async function down() {
  console.log('We are down!');
  return Promise.delay(2500);
}
```

This will translate to an object with functions for the `up` and `down` keys, which is the desired outcome.

jukkah:
There is now better way to do this: [import() operator](https://github.com/tc39/proposal-dynamic-import). It is now in stage 3 so its API is already frozen. That means, this will be more forward compatible than Babel's output.

I've made a lot of refactoring in past so I implemented this on behalf of you in #132.

PR #99, opened 2016-11-03
Added option timestamps to sequelize storage

Americas:
This PR adds the possibility to call the sequelize storage with timestamps set to true. This gives the migration table the `createdAt´ and ´updatedAt´ attributes.

This option defaults to false so there is no break in the logic.


axelhzf:
👍

jukkah:
Looks good. Merging.

PR #106, opened 2016-12-13
Feature: allows multiple different paths to sequelize migration folders to be given

renewooller:
This is useful for when models are shared between projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 .lock-wscript
 
 test/tmp
+
+# User's IDE data
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ node_modules
 .lock-wscript
 
 test/tmp
+test/tmp2
 
 # User's IDE data
 .vscode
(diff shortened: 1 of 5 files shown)


renewooller:
Shame we can't use arrow functions hey!

PascalPflaum:
@renewooller 

Thanks for your contribution.
Would you mind resolving those conflicts and test the changes again against the current codebase?

It would also be nice, if you would extend the readme to make this new feature transparent.

jedwards1211:
I'd like to have the ability to just pass in the list of migration files myself.  I'm looking for a way to make it easy for plugins in my app to add their own migrations to the list of core migrations, and it would be easy enough for me to just do all the globbing, concatenate the arrays of files, and pass it to umzug.

renewooller:
So as to 'fix-forward' I thought I'd wait until we're on a version that supports arrow functions.

tamlyn:
This is obsoleted by #164, right?

jedwards1211:
@tamlyn yeah #164 is not the most convenient really, it just gives the user the most possible control.

jedwards1211:
Also useful for when one wants plugins to an app to provide their own extra migrations

PR #108, opened 2017-01-19
fix(pattern-validation-error): Throw error when a file is not matching the pattern

yannvr:
It's a trivial logging addition to disclose when a migration task file pattern is not matching the default one. It seems to be a common gotcha that would save a lot of time to Umzug adopters because invalid file patterns are silently ignored.


jukkah:
I've made much refactoring in the code base or at least git thinks so. To help you, I rewrite this in 890e5e1 so that you don't need to update your code.

PR #115, opened 2017-03-02
Reject migration if umzug can't find the migration method

aboveyou00:
Rationale: you will never have a no-op migration. Currently, if the up or down methods can't be found when running a migration, Umzug replaces the method body with Promise.resolve() and tells you that the migration was run correctly. (It wasn't.)

This PR rejects the migration if you try to run it but that function doesn't exist.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -97,7 +97,8 @@ module.exports = redefine.Class(/** @lends Migration.prototype */{
    */
   _exec: function (method, args) {
     var migration  = this.migration();
-    var fun        = migration[method] || Bluebird.resolve;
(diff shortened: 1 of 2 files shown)


crrobinson14:
My contributions to this code block were tiny, just swept in from https://github.com/sequelize/umzug/pull/76. But I heartily endorse this contribution. ;)

aboveyou00:
Do I need to create an issue in order to get this pulled in? I couldn't find any details on contributing, but I know different repositories have different processes...

jukkah:
LGTM, merging :+1:

PR #132, opened 2017-04-26
Feature: Support ES6 modules and default exports

jukkah:
Add support to ES6 modules. It actually becomes useful only when Node.js gets
support for ES6 modules in the future.

This implementation supports both default exported object and direct named
exports.

```js
// case 1: default exported object
export default {
  up() {},
  down() {},
}

// case 2: named exports
export function up() {}
export function down() {}
```

# BREAKING CHANGES
- ~`Migration.migration()`, `Migration.up()`, and `Migration.down()` are now async.
  They are not in public API but potentially used from external apps.~
- `Migration.migration()` returns now `Promise<Object>` not `Object`. It is not in public API but potentially used from external apps.
- `Migration.up()` and `Migration.down()` returns now `Promise` not `*|Promise`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,6 +7,9 @@
     }]
   ],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "syntax-dynamic-import",
+    "dynamic-import-node",
+    "transform-runtime"
   ]
 }
--- a/src/migration.js
+++ b/src/migration.js
@@ -38,9 +38,9 @@ module.exports = class Migration {
    * Tries to require migration module. CoffeeScript support requires
    * 'coffee-script' to be installed.
    *
-   * @returns {Object} Required migration module
+   * @returns {Promise.<Object>} Required migration module
    */
-  migration() {
+  async migration() {
     if (this.path.match(/\.coffee$/)) {
       // 1.7.x compiler registration
       helper.resolve('coffee-script/register') ||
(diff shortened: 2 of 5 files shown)


src/migration.js
```diff
@@ -54,24 +54,24 @@ module.exports = class Migration {
...
 
   /**
    * Executes method `down` of migration.
    *
-   * @returns {*|Promise}
+   * @returns {Promise}
    */
-  down() {
+  async down() {
     return this._exec(this.options.downName, [].slice.apply(arguments));
```
aboveyou00: There is no reason to make the "up" and "down" methods async if you aren't going to await anything.

_exec still returns a promise, so you could either remove async from the signature, or you could await "this._exec(...)".

src/migration.js
```diff
@@ -89,14 +89,17 @@ module.exports = class Migration {
...
+    const migration = await this.migration();
+    let fun = migration[method];
+    if (migration.default) {
+      fun = migration.default[method] || migration[method];
+    }
     if (!fun) return Bluebird.reject('Could not find migration method: ' + method);
-    var wrappedFun = this.options.migrations.wrap(fun);
+    const wrappedFun = this.options.migrations.wrap(fun);
 
     return wrappedFun.apply(migration, args);
```
aboveyou00: Here you're mixing promise rejections and async/await. If you want to reject inside of an async function, just throw an exception - it translates to the same thing.

Also, if wrappedFun returns a promise (and I believe it does)  you should await that as well, rather than simply returning it.

src/migration.js
```diff
@@ -54,24 +54,24 @@ module.exports = class Migration {
       })();
     }
 
-    return require(this.path);
+    return import(this.path);
```
aboveyou00: If you aren't awaiting anything, don't mark the method as async. (Unless import actually returns Object and you are depending on the compiler to wrap it in a Promise.resolve())

.babelrc
```diff
@@ -7,6 +7,9 @@
     }]
   ],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "syntax-dynamic-import",
+    "dynamic-import-node",
+    "transform-runtime"
```
aboveyou00: I don't know how to use babel, so I can't really review this part of the PR.

src/migration.js
jukkah: Thanks, this needs some refactoring. You're thinking about lines 101 and 104? I'll change these two lines.

PR #147, opened 2017-08-31
MongoDB Storage

PascalPflaum:
Adds a mongoDB storage.
I used and commited to https://github.com/patrickhulce/umzug-mongo in the past, but that is no longer maintained so I would like to see this functionality here.


src/storages/MongoDBStorage.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * @class JSONStorage
+ */
+export default class MongoDBStorage extends Storage {
+    /**
+     * Constructs MongoDB file storage.
+     *
+     * @param {Object} [options]
+     * @param {String} [options.path='./umzug.json'] - Path to JSON file where
```
jukkah: Missing documentation for options.connection, options.collection, and options.collectionName.

PascalPflaum:
done. Please review again.

Issue #146, opened 2017-08-29
Gulp lint target not running


PascalPflaum:
I'm writing a MongoDB Storage (https://github.com/PascalPflaum/umzug/tree/feature/MongoDBStorage) right now and noticed that the gulp lint target is not working with the current code. 

>
src/index.js
  line 53  col 7   Expected '}' to match '{' from line 47 and instead saw '...'.
  line 53  col 10  Missing semicolon.
  line 54  col 5   Unexpected '}'.
  line 53  col 17  Expected an assignment or function call and instead saw an expression.
  line 53  col 18  Missing semicolon.
  line 54  col 6   Unnecessary semicolon.
  line 56  col 9   Expected an identifier and instead saw 'this' (a reserved word).
  line 56  col 13  Expected ')' to match '(' from line 56 and instead saw '.'.
  line 56  col 13  'function closure expressions' is only available in Mozilla JavaScript extensions (use moz option).
  line 56  col 68  Unexpected ')'.
  line 56  col 70  Unexpected '{'.
  line 57  col 13  Class properties must be methods. Expected '(' but instead saw 'new'.
  line 57  col 23  Expected an identifier and instead saw 'The logging-option should be either a function or false'.
  line 57  col 80  'function closure expressions' is only available in Mozilla JavaScript extensions (use moz option).
  line 57  col 81  Expected an identifier and instead saw ';'.
  line 58  col 6   Missing semicolon.
  line 64  col 23  Missing "use strict" statement.
  line 65  col 7   Expected '}' to match '{' from line 60 and instead saw '...'.
  line 65  col 10  Missing semicolon.
  line 66  col 5   Unexpected '}'.
  line 65  col 33  Expected an assignment or function call and instead saw an expression.
  line 65  col 34  Missing semicolon.
  line 65  col 33  Unrecoverable syntax error. (10% scanned).

  ✖  12 errors
  ⚠  11 warnings

If you need help fixing this issues, I would do so in a seperate branch.

PascalPflaum:
I suggest to switch to eslint instead eshint. Eshint doesn't support ES7 and spread operator not yet AFAIK.

jukkah:
PR is more than welcome +1

PascalPflaum:
adressed in #148

PR #148, opened 2017-09-08
Lint

PascalPflaum:
Repaired the gulp target by replacing eshint with eslint.

I tried to find a good ruleset for the current code. I took [StandardJS](https://standardjs.com) with some minor adjustments to those rules. For many rules there were different styles in the repo (like comma-dangle), so I took the setting value based on my personal opinion in those cases.


PascalPflaum:
Please notice that this branch includes #147.

src/index.js
```diff
@@ -7,13 +7,14 @@ import { EventEmitter } from 'events';
...
 import JSONStorage from './storages/JSONStorage';
+import MongoDBStorage from './storages/MongoDBStorage';
 import SequelizeStorage from './storages/SequelizeStorage';
 
 /**
  * @class Umzug
  * @extends EventEmitter
  */
-module.exports = class Umzug extends EventEmitter {
+export default class Umzug extends EventEmitter {
```
jukkah: Double check that this is actually working? I don't want to broke the package again.

PascalPflaum: You are right. So used to import/export that I totally forgot about classic require.
I will check it and will try to add a way to automatic test this. First idea I came up with is splitting up tests in unit tests (running against src) and integration tests (running against lib).

jukkah:
Ref #146

PR #149, opened 2017-09-09
feat: custom resolvers

mmkal:
@jukkah @PascalPflaum (not sure who's maintaining this repo atm).

This provides a way to run migrations on files other than JavaScript, by providing a custom resolver.

The custom resolver takes a file path as input, and returns an object of the form
```js
{ up: <function>, down: <function> }
```
i.e. the same shape as what we expect from `module.exports` for a regular js file.

There are a few examples in the `test/fixtures` folder - specifically: raw sql files, TypeScript, and CoffeeScript migrations.

I've recently gone through moving a project from [flyway](https://flywaydb.org/documentation/commandline/migrate) migrations to sequelize, and since my migrations were already sql files, I would have liked to be able to just point umzug/sequelize-cli to my existing migrations folder. The other justification is that it's a much more flexible way of achieving the CoffeeScript support that's already 'manually' implemented. TypeScript for example is now much more prevalent than CoffeeScript so it didn't make a lot of sense to me that CoffeeScript migrations were supported but TypeScript wasn't. This implementation allows for *any* compile-to-js language.

IMO, if this gets in, it would be a good idea to deprecate the explicit CoffeeScript support in favour of this, possibly replacing it with out-of-the-box `customResolver`s for CoffeeScript/TypeScript.


mmkal:
@jukkah @PascalPflaum ?

PascalPflaum:
Hi.

Thanks for the contribution.
I like the idea and will have a deeper look tomorrow.

PascalPflaum:
I had now a closer look and can't find any issues.

There is just one thing I would like to ask for - A extension to the readme explaining the feature.

tvoloshyn:
Hi everyone,

this doesn't seems to be available through NPM yet.
I've done `npm install umzug` (got version `2.0.1`) and it doesn't seems to include code for custom resolvers (checked my `node_modules/umzug/lib`).
Thanks for great library :)

PascalPflaum:
Hi @tvoloshyn. You are right. Those changes are already in the master, but were not released to npm yet. Right now I'm maintaining the GitHub Repo, which was the reason those code were finally merged. 
But I don't have rights to publish on npm yet. - I'm in touch with @jukkah, who has the permission and I'm confident that we will get this change together with #80 on the road as 2.1.0 soon.

PR #164, opened 2018-02-12
support passing an array of Migrations

jedwards1211:
I want to rock the boat a bit: the current method of giving sequelize a directory and glob pattern to find migrations isn't very flexible, and neither are other options people have suggested (providing multiple directories, passing in an array of files).

I need to provide a way for plugins in my app to inject their own migrations that live in a different directory from the app core, and possible even use a mix of languages (coffee, typescript, etc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,7 @@ It is possible to configure *umzug* instance via passing an object to the constr
   // The name of the negative method in migrations.
   downName: 'down',
 
+  // (advanced) you can pass an array of Migration instances instead of the options below
   migrations: {
     // The params that gets passed to the migrations.
     // Might be an array or a synchronous function which returns an array.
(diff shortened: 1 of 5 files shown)


PascalPflaum:
thx, I will work through some more open PRs/Issues and then release a new version.

cjpark:
@jedwards1211 this is exactly what I need, except for one issue: the `Migrations` class isn't exported.

How are you getting at it? Do you do something hacky like directly `require`-ing `migrations.js`?

jedwards1211:
@cjpark ah right.  Yes, I was going to do it the hacky way like that.  Though I ended up making code to copy migrations from various different folders into one temporary folder that I could use with Umzug as it is.  (This is all to make our server support plugins that can provide their own database models and migrations)

odedniv:
Is this repo still getting releases? I want this :smile:

PR #165, opened 2018-02-19
Update README.md with respect to MongoDBStorage

muellermarkus:
Just two small fixes: 

In the description paragraph for `MongoDBStorage`, `SequelizeStorage` is wrongly mentioned (although it links to the correct file). 

Additionally, `mongodb` and `none` are not mentioned as a possible storages in the code example.


PascalPflaum:
thx! Merged!

PR #169, opened 2018-04-17
feat: support passing array of migration options

### LINKED ISSUE #151: Support for multiple migrations paths in Umzug options
Hello, is there any possibility to define multiple directories in Umzug migration.path option?
For example:
```js
const umzug = new Umzug({
  migrations: {
    path: ['./dir1', './dir2'],
  }
})
```
I see that there is one rejected PR which implements this feature.

tamlyn:
`options.migrations` can now be an array of options objects:
- allows users to load data from multiple different folders (fixes #151 and supersedes #106)
- has use cases like specifying a resolver only for a certain file type
- still supports passing an array of `Migration`s in `options.migrations` (#164) but this is deprecated in favour of `options.migrationInstances`.
- doesn't change any public API but does change the shape of the options passed to `Migration`

Wasn't sure which file to put the new test in so created a new file.

Didn't know if I should update the Changelog or the version number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,8 @@ It is possible to configure *umzug* instance by passing an object to the constru
   // The name of the negative method in migrations.
   downName: 'down',
 
-  // (advanced) you can pass an array of Migration instances instead of the options below
+  // This can be an object or an array of objects. Pass an array if you want to 
+  // load from different paths, or have a different resolver for different file types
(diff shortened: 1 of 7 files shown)


tamlyn:
*bump* - anything I can do to help get this merged? Should I try and break it into smaller PRs?

PR #172, opened 2018-05-29
adding redis as storage

jnssthl:
Added a storage class to store executed migrations using Redis.


papb:
Hello, this is better as a separate module.

PR #188, opened 2019-01-14
feat: expose Migration class and document as public API

### LINKED ISSUE #184: Array of Migration instances?
Is there an example of passing an array of migration instances to the configuration object?  I see in the documentation:

`// (advanced) you can pass an array of Migration instances instead of the options below`

but I'm not sure how that would work.  I've tried:

```
const migrations = [
  { up: function () { ... }, down: function () { ... }
]

const umzug = new Umzug({
  ...,
  migrations: migrations
})
```

but that doesn't seem to be correct.

jedwards1211:
fix #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ The a subset of the `Umzug` constructor options
 
 The name of the migrate method
 
-##### `options.downName` (optional, default: `'up'`)
+##### `options.downName` (optional, default: `'down'`)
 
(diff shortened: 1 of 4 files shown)


starker-xp:
I found one error in your README.md

`##### `options.downName` (optional, default: `'up'`)`
you meant
`##### `options.downName` (optional, default: `'down'`)`

Issue #182, opened 2018-11-20
Ignoring sourcemaps


dorrotman:
Hello,

I am using TypeScript, and the compiler generates sourcemap files alongside all javascript files.
Taking into consideration this commit: https://github.com/sequelize/umzug/commit/890e5e16983fa52a258740d0c5246bc97a58d432
This causes warnings to be logged for each sourcemap file:
````
....js.map does not match pattern: ...
````
The commit description says:

> If there is other files, the pattern is probably filtering out some migrations.

However, in my opinion, sourcemaps are very common and should be allowed to exist alongside the javascript files. Come to think of it, the reason I use a filter is **because** I have other files and want to ignore them.

I would like to come up with a solution for this.
Possible solutions:

1. Ignoring sourcemap files before trying testing the `pattern` matching
2. Adding a `suppressWarnings` flag to the configuration so that it will not be logged
 

What do you think is the best solution?

Thanks!

ephemer:
> Come to think of it, the reason I use a filter is because I have other files and want to ignore them.

this.

Can we please get rid of these logs entirely?

po5i:
I'll be working on improving this verbosity level, I'm thinking of adding the new option, and probably a companion PR to sequelize-cli as well.

sonicdoe:
With https://github.com/sequelize/umzug/commit/cdfd53ee52bd7d9a64a6d0f11cb4c76f2ad3c6d3, non-migration files are no longer logged at all, so I think this issue can be closed.

sushantdhiman:
This should be fixed with `2.3.0` release

Issue #177, opened 2018-10-09
Log polution


blabno:
Please do not pollute logs with `File: does not match pattern`. If you really want to log it then move it to debug level.

sonicdoe:
With https://github.com/sequelize/umzug/commit/cdfd53ee52bd7d9a64a6d0f11cb4c76f2ad3c6d3, this log line has been removed completely.

sushantdhiman:
This should be fixed with `2.3.0` release

PR #189, opened 2019-02-12
Exclude patterns to avoid unnecessary logging

po5i:
Addresses #182 #177

This PR attempts to add a migration option called `exclude` which is an array of patterns. If any file that is not a migration file and compiles any of these patterns, it will be ignored, otherwise the usual log will show up.


po5i:
Hey @PascalPflaum would you mind to take a look or let me know who I should ping to in order to get this reviewed?

sonicdoe:
By the way, with https://github.com/sequelize/umzug/pull/190 being merged (which removes the log line altogether), I don’t think these changes are necessary anymore.

PR #190, opened 2019-04-09
don't log non-migration files

mmkal:
@PascalPflaum very minor thing, but it's been bugging me for a while. There's not much value in logging when a filename isn't matched by `pattern` since that's the whole reason `pattern` exists. It can end up a little noisy on stdout.


mmkal:
@SimonSchick @PascalPflaum is there a timeline for this to be released?

SimonSchick:
@mmkal I do not have publishing access.

PR #193, opened 2019-09-03
Support symlinked migrations directory

oreoluwa:
@jukkah I don't know how useful this would be, considering the outlined roadmap. 
However, I was in the same situation as #33 and from the thread, I observed it was now possible to `traverseDirectories` following #80 
This PR basically builds on that idea.

The idea here is:
An app that gets it's migrations from
```
/plugins
  | A/migrations
  | B/migrations
  | C/migrations
```
can easily just symlink the plugins into the main app migrations folder like:
```sh
ln -s plugins/A app/migrations/A
...
```
And configure sequelize to traverse directories for migrations.

*Why change from `lstatSync` to `statSync`?*
https://github.com/nodejs/node/issues/25342 and https://nodejs.org/api/fs.html#fs_fs_lstat_path_options_callback
Basically, 
```js
fs.lstatSync(filePath).isDirectory() => false
fs.lstatSync(filePath). isSymbolicLink() => true
```
returns the information about the link itself: which is `is the link a directory?` => No
While
```js
fs.statSync(filePath).isDirectory() => true
fs.statSync(filePath).isSymbolicLink() => false
```
gives information about the directory it's pointing at.

### DIFF
--- a/test/Umzug/up.test.js
+++ b/test/Umzug/up.test.js
@@ -304,3 +304,20 @@ describe('up-directories', () => {
 
   upTestuite();
 });
+
+describe('up-directories with symbolicRef', () => {
+  beforeEach(function () {
+    helper.clearTmp();
+    return helper
+      .prepareMigrations(3, { directories: [['1', '2'], ['1', '2'], ['1', '3', '4', '5']], usesSymbolicRef: true })
+      .then((migrationNames) => {
+        this.migrationNames = migrationNames;
+        this.umzug = new Umzug({
+          migrations: { path: join(__dirname, '/../tmp/1'), traverseDirectories: true },
+          storageOptions: { path: join(__dirname, '/../tmp/umzug.json') },
+        });
+      });
+  });
+
+  upTestuite();
+});
--- a/test/helper.js
+++ b/test/helper.js
@@ -7,10 +7,9 @@ const helper = module.exports = {
     const tmpPath = join(__dirname, '/tmp');
     path = path || tmpPath;
     const files = fs.readdirSync(path);
-
     files.forEach((file) => {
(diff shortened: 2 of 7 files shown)


test/helper.js
```diff
@@ -36,9 +37,18 @@ const helper = module.exports = {
...
+        const dirName = randomName();
+        migrationsPath = join(__dirname, '/tmp', dirName, '/');
+        path = join(path, dirName);
+        fs.mkdirSync(migrationsPath);
+        fs.symlinkSync(migrationsPath, path);
+      }
     }
     fs.writeFileSync(
-      join(path, name + '.js'),
+      join(migrationsPath, name + '.js'),
```
oreoluwa: observe the use of migrationsPath here, which could be pointing to any external directory or plugins

test/Umzug/up.test.js
```diff
@@ -304,3 +304,20 @@ describe('up-directories', () => {
...
+
+describe('up-directories with symbolicRef', () => {
+  beforeEach(function () {
+    helper.clearTmp();
+    return helper
+      .prepareMigrations(3, { directories: [['1', '2'], ['1', '2'], ['1', '3', '4', '5']], usesSymbolicRef: true })
+      .then((migrationNames) => {
+        this.migrationNames = migrationNames;
+        this.umzug = new Umzug({
+          migrations: { path: join(__dirname, '/../tmp/1'), traverseDirectories: true },
```
oreoluwa: observe that all migrations are being run from the `tmp/1` directory, which could be the `migrations` directory of the app.
Note that the actual migrations were created in multiple directories inside `./tmp`, i.e. `'./tmp' + randomName`.

PR #195, opened 2019-09-30
Rename `build` script to `prepare`

sonicdoe:
Enables Umzug to be installed from a Git repository. [From the npm documentation](https://docs.npmjs.com/cli/install):

> If the package being installed contains a `prepare` script, its `dependencies` and `devDependencies` will be installed, and the prepare script will be run, before the package is packaged and installed.


papb:
Hello! I see you are a first-time contributor, thank you for taking the time to help Sequelize! :)

PR #196, opened 2020-01-27
feat: add format function

jaulz:
Sometimes you want to modify the name of the migration (e.g. in my case I would like to remove the file suffix because it changes depending on whether I run the Typescript or compiled version). This PR introduces a format function in the options that takes care of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,9 +150,9 @@ const umzug = new Umzug({
         return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) };
     }
 
-    // A function that receives the file name of the migration and returns the name of the 
+    // A function that receives the file path of the migration and returns the name of the 
     // migration. This can be used to remove file extensions for example.
-    nameFormatter: function (fileName) {
+    nameFormatter: function (filePath) {
         return path.parse(fileName).name;
     }
   }
--- a/README.md
+++ b/README.md
@@ -349,8 +349,14 @@ It is possible to configure *umzug* instance by passing an object to the constru
     // read raw sql etc.
     // See https://github.com/sequelize/umzug/tree/master/test/fixtures
     // for examples.
-    customResolver: function (sqlPath)  {
-        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) }
+    customResolver: function (sqlPath) {
+        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) };
+    }
+
+    // A function that receives the file name of the migration and returns the name of the 
+    // migration. This can be used to remove file extensions for example.
+    nameFormatter: function (fileName) {
+        return path.parse(fileName).name;
     }
   }
 }
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ const umzug = new Umzug({
     // A function that receives the file path of the migration and returns the name of the 
     // migration. This can be used to remove file extensions for example.
     nameFormatter: function (filePath) {
-        return path.parse(fileName).name;
+        return path.parse(filePath).name;
     }
   }
 })
--- a/src/index.js
+++ b/src/index.js
@@ -44,6 +44,9 @@ module.exports = class Umzug extends EventEmitter {
    * function that specifies how to get a migration object from a path. This
    * should return an object of the form { up: Function, down: Function }.
    * Without this defined, a regular javascript import will be performed.
+   * @param {Migration~format} [options.migrations.format] - A function that
+   * receives the file name of the migration and returns the name of the 
+   * migration. This can be used to remove file extensions for example.
    * @constructs Umzug
    */
   constructor (options = {}) {
@@ -69,6 +72,7 @@ module.exports = class Umzug extends EventEmitter {
         pattern: /^\d+[\w-]+\.js$/,
         traverseDirectories: false,
         wrap: fun => fun,
+        format: undefined,
         ...this.options.migrations,
       };
     }
@@ -154,7 +158,7 @@ module.exports = class Umzug extends EventEmitter {
    * @returns {Promise.<Migration>}
    */
   executed () {
-    return Bluebird.resolve(this.storage.executed()).bind(this).map((file) => new Migration(file));
+    return Bluebird.resolve(this.storage.executed()).bind(this).map((file) => new Migration(file, this.options));
   }
 
   /**
--- a/src/index.js
+++ b/src/index.js
@@ -46,7 +46,7 @@ module.exports = class Umzug extends EventEmitter {
    * should return an object of the form { up: Function, down: Function }.
    * Without this defined, a regular javascript import will be performed.
    * @param {Migration~nameFormatter} [options.migrations.nameFormatter] - A
-   * function that receives the file name of the migration and returns the name
+   * function that receives the file path of the migration and returns the name
    * of the migration. This can be used to remove file extensions for example.
    * @constructs Umzug
    */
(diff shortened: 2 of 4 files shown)


kf6kjg:
Watch your whitespace, that's what's causing the CI to fail.  Otherwise this looks good to me and would allow me to solve the same problem.

However I'd go one step further and provide a way to format the name that is in storage too.  That way I can take an existing database that was migrated without a formatter and still not have a migration collision.  You could use the same format function, it would just force the user to add additional logic.  Or you can have different functions for each side: one for formatting files from disk, another for formatting migration names from storage.

src/migration.js
```diff
@@ -29,12 +29,20 @@ module.exports = class Migration {
...
+   * receives the file name of the migration and returns the name of the 
+   * migration. This can be used to remove file extensions for example.
    * @constructs Migration
    */
   constructor (path, options) {
     this.path = _path.resolve(path);
-    this.file = _path.basename(this.path);
     this.options = options;
+
+    if (options && typeof options.migrations.format === 'function') {
```
papb:
```suggestion
    if (options && options.migrations && typeof options.migrations.format === 'function') {
```

src/migration.js
```diff
@@ -29,12 +29,20 @@ module.exports = class Migration {
...
+        this.file = options.migrations.format(this.path);
```
papb:
```suggestion
      this.file = options.migrations.format(this.path);
      if (typeof this.file !== 'string') {
        throw new Error(`Unexpected migration formatter result for '${this.path}': expected string, got ${typeof this.file}`);
      }
```

src/migration.js
```diff
@@ -29,12 +29,20 @@ module.exports = class Migration {
...
+    } else {
+        this.file = _path.basename(path);
```
papb:
```suggestion
      this.file = _path.basename(path);
```

src/index.js
```diff
@@ -44,6 +44,9 @@ module.exports = class Umzug extends EventEmitter {
    * function that specifies how to get a migration object from a path. This
    * should return an object of the form { up: Function, down: Function }.
    * Without this defined, a regular javascript import will be performed.
+   * @param {Migration~format} [options.migrations.format] - A function that
```
papb: Please rename this new parameter from `format` to `nameFormatter`

papb:
Hi @jaulz, thanks for the PR! Please address my comments and also address the suggestion by @kf6kjg. Thanks!

jaulz:
Thanks for your input! I just pushed the updated files. 

@kf6kjg do you mean something like this?
https://github.com/sequelize/umzug/pull/196/files#diff-410142b027abfe35d3351577b731ebcaR106-R107

kf6kjg:
Yeah, that looks like it should work.  I like it.  It'd be good to give and example of how to use the option in the documentation such as

```javascript
{
  nameFormatter: path => {
    const filename = Path.basename(path); // Isolate the file name from the full path.
    return filename.slice(0, filename.lastIndexOf(".")); // Remove the file extension, leaving just the unique name of the migration.
  }
}
```

At least that's what I'd use as I have migrations being run under Node on the live server at the moment and I'd like to eventually change over to using Node-TS.  This would prevent the extension change, .js to .ts, from causing any headaches.

kf6kjg:
Though, reviewing the code a little, I think it'd be good for you to write up some unit tests to exercise this. Also to make sure to fix whatever's causing the existing unit tests to fail - but I suspect you are already on top of that! :D

jaulz:
@kf6kjg okay, it's ready 😊

test/fixtures/index.js
```diff
@@ -32,6 +32,7 @@ describe('custom resolver', () => {
           ],
           pattern: this.pattern,
           customResolver: this.customResolver,
+          nameFormatter: (path) => basename(path).replace(/\.[^/.]+$/, ''), // Remove file extension
```
papb: Please use [`path.parse().name`](https://nodejs.org/api/path.html#path_path_parse_path) instead :)

jaulz: Ah, thanks for the tip!

README.md
```diff
@@ -146,8 +146,14 @@ const umzug = new Umzug({
     // read raw sql etc.
     // See https://github.com/sequelize/umzug/tree/master/test/fixtures
     // for examples.
-    customResolver: function (sqlPath)  {
-        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) }
+    customResolver: function (sqlPath) {
+        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) };
```
jaulz: I added a semicolon here to keep it consistent.

jaulz:
Okay, I added it including the example of the file extension removal.

README.md
```diff
@@ -349,8 +349,14 @@ It is possible to configure *umzug* instance by passing an object to the constru
...
+    }
+
+    // A function that receives the file name of the migration and returns the name of the 
+    // migration. This can be used to remove file extensions for example.
+    nameFormatter: function (fileName) {
```
papb: Does it receive the file name or the absolute path?

jaulz: I adjusted the functions and readme.

README.md
```diff
@@ -146,8 +146,14 @@ const umzug = new Umzug({
...
-    customResolver: function (sqlPath)  {
-        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) }
+    customResolver: function (sqlPath) {
+        return { up: () => sequelize.query(require('fs').readFileSync(sqlPath, 'utf8')) };
+    }
+
+    // A function that receives the file path of the migration and returns the name of the 
+    // migration. This can be used to remove file extensions for example.
+    nameFormatter: function (filePath) {
+        return path.parse(fileName).name;
```
papb:
```suggestion
        return path.parse(filePath).name;
```

src/migration.js
```diff
@@ -92,7 +105,8 @@ module.exports = class Migration {
    * @returns {boolean}
    */
   testFileName (needle) {
-    return this.file.indexOf(needle) === 0;
+    const formattedNeedle = this.options.migrations.nameFormatter(needle);
+    return this.file.indexOf(formattedNeedle) === 0;
```
papb: Are you sure calling `nameFormatter` here is necessary?

jaulz: That was the idea of @kf6kjg in his first response:
> However I'd go one step further and provide a way to format the name that is in storage too. That way I can take an existing database that was migrated without a formatter and still not have a migration collision. You could use the same format function, it would just force the user to add additional logic. Or you can have different functions for each side: one for formatting files from disk, another for formatting migration names from storage.

papb: Hmm... But is this `needle` (which I think is a terrible variable name by the way - not your fault of course) a full path? Or just a name? Because the `nameFormatter` function expects an absolute path as argument, right?

jaulz: `needle` is in this case the file name which was previously used for the migration but you can remove extensions etc. of course as well in the `nameFormatter`.

kf6kjg: I'm OK with this section of the code: a file name IS a relative path, thus any parser should behave itself.  My caveats are going to be in a different comment layer.

test/fixtures/index.js
```diff
@@ -69,13 +76,15 @@ describe('custom resolver', () => {
           downName: 'down',
           migrations: {
             wrap: fn => () => fn(this.sequelize.getQueryInterface(), this.sequelize.constructor),
+            nameFormatter: (path) => parse(path).name,
```
papb: Question: Is this addition to this test necessary for it to pass? Would it fail without this?

jaulz: Yes, it would because then this check would fail as the file extensions are missing:
https://github.com/sequelize/umzug/blob/0cf1ee75dc6375647772d6ca5010b7fedacd2af2/test/fixtures/index.js#L56

evantahler:
This is excellent! I've got the same use-case (ts vs js file extensions).  
I don't see this in the master branch yet?

PR #199, opened 2020-03-06
Added migrationsList helper to easily build a valid list of migrations

nicolas-van:
Hello,

This PR brings a solution to a problem that has already been reported, notably in this issue: https://github.com/sequelize/umzug/issues/184 . This older PR also attempts to bring a solution for that: https://github.com/sequelize/umzug/pull/188 .

This issue is that the documentation specifies that it is possible to avoid using an automatic file loader for migrations and to specify a list of migrations instead. (Notable use case where we need to do this is when we use Webpack, where a file loader usually won't work as expected.) Unfortunately this is not particularly easy to do. The `Migration` class is not exposed in the `index.js` file and, even by importing it anyway, it's still not easy to use because it is very oriented toward that migration-file-loading behavior.

This PR simply adds an utility function named `migrationsList()` that allow to specify a list of migrations this way:

```js
const umzug = new Umzug({
  migrations: Umzug.migrationsList([
    {
      name: '00-first-migration',
      up: ...,
      down: ...
    }
  ], 
  // a facultative list of parameters that will be sent to the `up` and `down` functions
  [sequelize.getQueryInterface()])
})
```

I added a test and documented the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ const umzug = new Umzug({
   // The name of the negative method in migrations.
   downName: 'down',
 
-  // (advanced) you can pass an array of Migration instances instead of the options below
+  // (advanced) you can pass an array of migrations built with `migrationsList()` instead of the options below
   migrations: {
     // The params that gets passed to the migrations.
     // Might be an array or a synchronous function which returns an array.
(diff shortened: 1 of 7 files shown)


nicolas-van:
Thank you.

PR #202, opened 2020-03-22
Configure babel to not rely on babel-runtime

papandreou:
Yields a 123 bytes increase in the transpiled files, but avoids 8 MB of dependencies 😌


sushantdhiman:
Nice

tgabi333:
:+1:

PR #206, opened 2020-03-29
Super refactor

papb:
* Drop support for Node <10
* Drop babel and coffee-script from devDependencies
* Several other improvements


papb:
Note: removing Babel entirely might look like a breaking change but it's not really. Although we did have tests which were supposed to cover the ability of umzug to operate on JS files using the ES6 module syntax, these tests were actually misleading because they only worked since we happened to have `@babel/register` in our own `devDependencies`.

papb:
Also `coffee-script` support was dropped because this "support" was actually built upon [a npm package that has been deprecated](https://www.npmjs.com/package/coffee-script) and has no releases for three years. It was a small price to pay to allow easier refactoring that will be coming soon. Thanks for understanding!

PR #208, opened 2020-04-01
Support custom sorting function

### LINKED ISSUE #153: custom sorting function
Umzug sorts migration filenames by name, thus files named `1`, `2`, `12` will be sorted `1`, `12`, `2`. Files can also be named using different paterns (`major.minor` etc.), so it's best to allow to supply custom sorting function.

rockers7414:
Add the `migrationSorting` argument to support custom sorting. (resolve #153)


papb:
Hello! Thanks for the PR!

I have merged & published a huge rewrite so your PR is now outdated from the codebase :sweat_smile: can you please rebase?

In short, you will have to re-implement what you did into `umzug.ts` now.

Thank you!

rockers7414:
It's really a huge change. I have re-implemented with typescript. Thanks!

papb:
@rockers7414 Thanks for rebasing! May I ask you to please create a test for this feature, in `test/test.ts`?

Hint: for quicker testing you can run `npx ava test/test.ts` instead of `npm test`.

rockers7414:
@papb Hey, I've created the test for PR. Please help to review it. Thanks!

src/umzug.ts
```diff
@@ -375,7 +388,7 @@ export class Umzug extends EventEmitter {
 	*/
 	private async _findMigrations(): Promise<Migration[]> {
 		if (Array.isArray(this.options.migrations)) {
-			return this.options.migrations;
+			return this.options.migrations.sort((a, b) => this.options.migrationSorting(a.file, b.file));
```
papb, line 391: Since `sort` operates in-place, this changes the value of `this.options.migrations` - is this intended? If not, please use `.slice().sort()` instead.

rockers7414: In fact, I do want to change the order of the value of `this.options.migrations`. In my case, I'm using the version as the filename, such as `1.0.0`. So I want to make it run the migration script with the correct order.

Is there any issue or side effect if we change the value of `this.options.migrations`?

papb: I don't think altering it directly is a good idea, I wouldn't expect a class to modify a value I passed into it. In the future I hope to enable the [prefer-readonly-parameter-types rule](https://github.com/sequelize/umzug/blob/69d7b571e66680b93218599a868f824969daefb6/.eslintrc.js#L55).

Please try to apply your change in [this line](https://github.com/sequelize/umzug/blob/fe2bb23b2499d750fa6461a7487e15644fffe359/src/umzug.ts#L32) instead, using `.slice().sort()`. Thanks!

rockers7414: Fixed. Thanks.

papb:
[Released at v3.0.0-beta.4](https://github.com/sequelize/umzug/releases/tag/v3.0.0-beta.4)!

PR #213, opened 2020-04-19
chore(lint): prettier

mmkal:
This adds prettier to the project and fixes a few rule violations. We're still using `xo`, but now as a regular eslint config, to avoid having too many places that our linter is configured, and to manage conflicts between tools with less complexity (prettier vs eslint vs typescript-eslint vs xo) - everything is now in .eslintrc.js, except for legacy-tests which have the old `.eslintrc.json`.

Most of the changes are just whitespace, you can view without line-ending fixes, tabs etc. here: https://github.com/sequelize/umzug/pull/213/files?w=1

This also adds a .gitattributes file to fix LF/CRLF inconsistencies.


src/types.ts
```diff
@@ -1,165 +1,153 @@
...
+	| {
+			/** List of migrations to execute (by name). */
+			migrations?: string[];
+
+			/** The migration to start executing from. The given migration is not executed itself, only the ones after it. */
+			from?: string;
+
+			/** The last migration to execute (included). */
+			to?: string;
+	  };
```
papb, line 153: There is something odd with indentation here. Also last line has a tab followed by 2 spaces, is mixing tabs and spaces intentional?

.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
...
+  "files.watcherExclude": {
+    "**/.git/objects/**": true,
+    "**/.git/subtree-cache/**": true,
+    "**/node_modules/**": true
+  },
+  "javascript.implicitProjectConfig.checkJs": true,
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  }
+}
```
papb, line 15: Can you please explain the point of each of these configs? And why should we enforce vscode settings on contributors?

legacy-tests/.eslintrc.json
```diff
@@ -10,6 +10,7 @@
   "rules": {
     "mocha/no-exclusive-tests": "error",
     "no-unused-expressions" : "off",
-    "no-new" : "off"
+    "no-new" : "off",
+    "prettier/prettier": "off"
```
papb, line 14: What does this do? (sorry have never used prettier)

package.json
```diff
@@ -20,30 +20,39 @@
 	},
 	"devDependencies": {
 		"@ava/typescript": "^1.1.1",
+		"@typescript-eslint/eslint-plugin": "^2.28.0",
+		"@typescript-eslint/parser": "^2.28.0",
 		"ava": "^3.5.0",
 		"chai": "^4.2.0",
 		"del-cli": "^3.0.0",
-		"eslint-config-xo-typescript": "~0.27.0",
```
papb, line 26: Why?

package.json
```diff
@@ -84,25 +93,5 @@
 			}
 		},
 		"timeout": "2m"
-	},
-	"xo": {
-		"extends": "xo-typescript",
-		"extensions": [
-			"ts"
-		],
-		"ignore": "legacy-tests",
```
papb, line 93: Is this still being ignored?

legacy-tests/.eslintrc.json
mmkal: Prettier auto-formats code - trailing commas, tabs, semicolons, etc. It's pretty opinionated though so on codebases that haven't been using a formatter turning it on can introduce a lot of whitespace-type changes. This here to avoid churn by not running on the legacy tests.

package.json
mmkal: Mistake - have reinstated. Good catch - thanks!

.vscode/settings.json
mmkal: These settings are some basics that make sure:

- the correct version of typescript is used (rather than the default which is the vscode global version. This can lead to works-on-my-machine type bug reports).
- _if_ the user has the eslint extension enabled, vscode will auto-fix fixable errors on save. It won't do anything for those who don't
- show the `.git` folder in the file explorer. It doesn't do much harm and is occasionally useful to look at.
- tells typescript not to try to watch node_modules and some other folders

Personally I prefer having these settings checked in as it makes for a more consistent developer experiences. Having said that, I realise this can be a bit of a controversial subject, so I've removed this for now - it can be discussed and merged/not-merged separately.

package.json
mmkal: No - the legacy tests have their own [eslint config](https://github.com/sequelize/umzug/blob/master/legacy-tests/.eslintrc.json) which is I believe pretty much what the project was using before your refactor. eslint is smart enough to handle hierarchical configs.

src/types.ts
mmkal: This is just prettier auto-formatting. In general, it doesn't allow you to decide on whitespace at all. I believe the reasoning for this mixed space/tab thing is because prettier is trying to align the closing brace with the opening one. The opening one has an extra indentation of two characters because of the `| `.

.vscode/settings.json
papb: Interesting, I like the idea, I think you can open a separate PR with this. Also what does `"javascript.implicitProjectConfig.checkJs": true` do?

legacy-tests/.eslintrc.json
papb: Ah, ok, so it's just disabling prettier entirely on the legacy tests :+1:

.eslintrc.js
```diff
@@ -0,0 +1,80 @@
...
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:unicorn/recommended',
+    'plugin:import/typescript',
+    'plugin:ava/recommended',
+    'xo',
+    'xo-typescript',
+  ],
```
papb, line 20: Interesting, why don't `xo` and `xo-typescript` have to be listed in the `plugins` option? And why they don't need the `plugin:` prefix here?

mmkal: `extends` is for configs - basically sets of rules; and `plugins` is for things that do pre-/post-processing on the code. See https://eslint.org/docs/developer-guide/working-with-plugins#configs-in-plugins

For the `plugin:` prefix: plugins can include optional configs, and the syntax for using them is `plugin:pluginName/configName`. As for `xo` and `xo-typescript` - those are the shorthand way of referring to the npm packages `eslint-config-xo` and `eslint-config-xo-typescript`.

PR #215, opened 2020-04-27
refactor: jest

mmkal:
This migrates from ava to jest. The PR keeps the existing ava tests/assertions exactly as they are, but as follow-ups, we can:

- use `jest.fn()` to make logging assertions (and drop `sinon` as a dependency)
- use inline snapshot testing
- stop compiling the tests - ts-jest does this in-memory. This can simplify our tsconfig build output from `lib/src/index.js` to `lib/index.js` etc.
- add vscode debugger configs (or document recommended configs)
- use [codemods](https://github.com/skovhus/jest-codemods) to migrate legacy-tests from mocha and into typescript (used in this PR to convert from ava)


papb:
I am in favor of all follow-ups you mentioned!!

P.S.: I will try to look at your PRs faster now. Sorry about that!

papb:
(by the way, he added a serial test, which is not ideal but I decided to accept anyway because a parallel test would require usage of another file for json storage, and I didn't want to add this extra "cognitive load" to that PR. If you feel like refactoring in order to make it parallel, please do, but it can wait to be done in the future as well)

papb:
Does jest run tests serially or in parallel? Since Umzug uses `./umzug.json` by default, testing in parallel might give unexpected results due to reuse of the same file. I like the fact that ava runs tests in parallel, I think it's a good way to ensure all tests are independent, how does jest work in this case?

mmkal:
Jest runs test files in parallel, but individual tests within the file serially: https://github.com/facebook/jest/issues/6957

To run tests within a file in parallel, there's `test.concurrent`, but it isn't widely used.

papb:
Ready to merge :rocket:

PR #217, opened 2020-05-04
test: convert storage tests to typescript

mmkal:
This moves the storage tests out of legacy-tests and into typescript/jest-land.

There are some changes I want to make to a few of the storage helpers, and how `Umzug` is passed them, so I wanted to get them tested with the new system first. I think some of them could be dropped and replaced with integration tests, long term:

- mongodb tests aren't testing all that much, since they mock out the actual MongoDB part
- same goes for sequelize. The tests are very complex and verbose. I didn't change anything about them other than converting to jest/typescript


test/storages/sequelize.test.ts
```diff
@@ -0,0 +1,408 @@
...
+					expect(description).toMatchInlineSnapshot(`
+						Object {
+						  "name": Object {
+						    "allowNull": false,
+						    "defaultValue": undefined,
+						    "primaryKey": true,
+						    "type": "VARCHAR(255)",
+						  },
+						}
+          `);
```
papb, line 57: Looks like you forgot to fix this last line in your `use tabs` commit (b840f0ec624438e9f1b00fda3dc2f7b35b1dd09f)

Also, by the way, I'm curious - does Jest automatically outdents the inline snapshot?

test/storages/sequelize.test.ts
```diff
@@ -0,0 +1,408 @@
...
+				});
+		});
+
+		it('returns an empty array if no migrations were logged yet', () => {
+			const storage = new Storage({
+				sequelize: helper.sequelize,
+			});
+
+			return storage.executed().then(migrations => {
+				expect(Object.keys(migrations)).toHaveLength(0);
```
papb, line 363: Just out of curiosity - wouldn't `expect(migrations).toEqual({})` work?

papb:
> I think some of them could be dropped and replaced with integration tests, long term: [...]

Agreed

test/storages/sequelize.test.ts
mmkal: Good catch - and yes jest is supposed to handle this automatically, but I think I need to tweak some of the lint config so it can tell that we like tabs over spaces. I'll do that in another PR I think and just fix this one manually.

test/storages/sequelize.test.ts
mmkal: Yes it probably would - this file is a straight up port, I just ran jest-codemods and cleaned up some of the context variables, I didn't touch any of the assertions.

PR #225, opened 2020-05-20
test: events

mmkal:
also decouple js from sql test fixtures

Follow-on from the discussion in a previous PR where we toyed with the idea of making the EventEmitter methods type declarations only. That'd work, but it occurred to me that we should probably have a test that makes sure the events are emitted at the right times.

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -166,7 +166,7 @@ test('events', async () => {
 		]),
 	});
 
-	// `.addListener` and `.on` are aliases -us both to make sure they're wired up properly
+	// `.addListener` and `.on` are aliases - use both to make sure they're wired up properly
 	umzug.addListener('migrating', spy('migrating'));
 	umzug.on('migrated', spy('migrated'));
 	umzug.addListener('reverting', spy('reverting'));
--- a/test/test.ts
+++ b/test/test.ts
@@ -156,7 +156,7 @@ test('migration sort', async () => {
 test('events', async () => {
 	const mock = jest.fn();
 	// eslint-disable-next-line @typescript-eslint/no-unsafe-return
-	const spy = (label: string) => (...args) => mock([label, ...args]);
+	const spy = (label: string) => (...args) => mock(label, ...args);
 
 	const umzug = new Umzug({
 		storage: memoryStorage(),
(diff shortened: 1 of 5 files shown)


test/test.ts
```diff
@@ -152,3 +152,45 @@ test('migration sort', async () => {
...
+
+	mock.mockClear();
+
+	await umzug.down();
+
+	expect(mock.mock.calls.join('\n')).toMatchInlineSnapshot(`
+		"reverting,m2,{migration: m2}
+		down
+		reverted,m2,{migration: m2}"
+	`);
```
papb, line 195: I think this mixture of

* `toMatchInlineSnapshot`;
* Overwriting `.toString`;
* Relying on the exact (implicit) stringification that happens behind the scenes in `.join`,

although understandable with some effort, is not very readable.

That said, I didn't have a better suggestion from the top of my head. However I was in the mood of jumping in Jest docs, and I discovered that `.toEqual` accepts nested matchers, which is fantastic (I think Jest has just won me now). So I came up with another way to write the test that I believe to be a lot more readable.

This is probably the largest `` ```suggestion``  I have ever given in a PR :sweat_smile: 

```suggestion
	const spy = (label: string) => (...args) => mock(label, ...args);

	const umzug = new Umzug({
		storage: memoryStorage(),
		migrations: migrationsList([
			{ name: 'm1', up: spy('up'), down: spy('down') },
			{ name: 'm2', up: spy('up'), down: spy('down') },
		]),
	});
	umzug.addListener('migrating', spy('migrating'));
	umzug.on('migrated', spy('migrated'));
	umzug.on('reverting', spy('reverting'));
	umzug.on('reverted', spy('reverted'));

	await umzug.up();

	expect(mock.mock.calls).toEqual([
		['migrating', 'm1', expect.objectContaining({ file: 'm1' })],
		['up'],
		['migrated', 'm1', expect.objectContaining({ file: 'm1' })],
		['migrating', 'm2', expect.objectContaining({ file: 'm2' })],
		['up'],
		['migrated', 'm2', expect.objectContaining({ file: 'm2' })],
	]);

	mock.mockClear();

	await umzug.down();

	expect(mock.mock.calls).toEqual([
		['reverting', 'm2', expect.objectContaining({ file: 'm2' })],
		['down'],
		['reverted', 'm2', expect.objectContaining({ file: 'm2' })],
	]);
```

papb: Also:

Why is `umzug.addListener('migrating', spy('migrating'))` needed, but not other `addListener` calls?

papb:
> also decouple js from sql test fixtures

Very good!

test/test.ts
mmkal: Good call-out. I use `.toMatchInlineSnapshot` because it's really fast to _write_. While building out a test it's a really good way to iterate fast, but often it's not a bad idea to switch to something a bit more set in stone like this once the test/feature is stable. I've gone for `toMatchObject` rather than `toEqual` - that way the `expect.objectContaining` part isn't even needed - `toMatchObject` basically applies the same rule deeply/recursively.

mmkal: I've added a comment explaining the use of both `addListener` and `on`. Since we've explicitly implemented both, I want the test to make sure we haven't screwed up either of them: https://github.com/sequelize/umzug/blob/2dcbd61da0c4e009a9e06ddf88b1663cbc527024/test/test.ts#L169

papb:
Excellent!!

PR #226, opened 2020-05-21
fix: workaround sequelize types in tests

mmkal:
One of the travis jobs runs `npm install sequelize@next`, which now breaks us. This gets things green again with a workaround. I've put in a todo that will become a lint error when we're on a stable version of sequelize using [expiring-todo-comments](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/expiring-todo-comments.md), to make sure we remember to clean up the workaround.

### DIFF
--- a/test/storage/sequelize.test.ts
+++ b/test/storage/sequelize.test.ts
@@ -9,7 +9,7 @@ import { v4 as uuid } from 'uuid';
 import jetpack = require('fs-jetpack');
 
 // TODO [sequelize@>=6.0.0] remove when sequelize fixes the types bug
-// a change in the types of sequelize@next makes `describe` a static function
+// a change in the types of sequelize@next makes `describe` a static function - see https://github.com/sequelize/sequelize/issues/12296
 // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
 // @ts-ignore
 const describeModel = (model: sequelize.Model) => model.describe();
--- a/test/storage/sequelize.test.ts
+++ b/test/storage/sequelize.test.ts
@@ -8,11 +8,9 @@ import { join } from 'path';
 import { v4 as uuid } from 'uuid';
 import jetpack = require('fs-jetpack');
 
-// TODO [sequelize@>=6.0.0] remove when sequelize fixes the types bug
-// a change in the types of sequelize@next makes `describe` a static function - see https://github.com/sequelize/sequelize/issues/12296
-// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-// @ts-ignore
-const describeModel = (model: sequelize.Model) => model.describe();
+// TODO [>=3.0.0]: Investigate whether we are mis-using `model.describe()` here, and get rid of `any`.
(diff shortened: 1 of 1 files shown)


papb:
Is there an issue in the main repo tracking this typing error? If not, can you please open one?

test/storage/sequelize.test.ts
```diff
@@ -8,6 +8,12 @@ import { join } from 'path';
 import { v4 as uuid } from 'uuid';
 import jetpack = require('fs-jetpack');
 
+// TODO [sequelize@>=6.0.0] remove when sequelize fixes the types bug
+// a change in the types of sequelize@next makes `describe` a static function
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+const describeModel = (model: sequelize.Model) => model.describe();
```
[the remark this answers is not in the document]
mmkal: I think ts-ignore is marginally better, since it doesn't lose type information with the current version of sequelize, which is what's installed by default.

mmkal:
> are you sure it's enabled?

Yes - I changed it to `>=5.0.0` locally and got an error in my IDE. I believe it comes in through the `unicorn` plugin, not xo: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/cec3a9dfa08db1efd166001d604b04d4fa8ab53f/index.js#L26

Re making an issue in the main repo - I'll try to find some time to get a repro that doesn't involve all the baggage of this test, but I don't actually use sequelize itself so might take a little while to get to it.

test/storage/sequelize.test.ts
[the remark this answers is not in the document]
papb: Ah, ok!

test/storage/sequelize.test.ts
```diff
@@ -8,6 +8,12 @@ import { join } from 'path';
 import { v4 as uuid } from 'uuid';
 import jetpack = require('fs-jetpack');
 
+// TODO [sequelize@>=6.0.0] remove when sequelize fixes the types bug
+// a change in the types of sequelize@next makes `describe` a static function - see https://github.com/sequelize/sequelize/issues/12296
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+const describeModel = (model: sequelize.Model) => model.describe();
```
papb, line 15:
```suggestion
const describeModel = (model: typeof sequelize.Model) => model.describe();
```

mmkal: This doesn't work because we don't have a `typeof Model`, we have a `Model` - this is called with the return value of `model.sync()` - from sequelize: https://github.com/sequelize/sequelize/blob/b2bccb8aa3faeb6a891fd9f7d3858ebff5b9c32f/types/lib/model.d.ts#L1604-L1608

I'm going to change to `any` for now. It might be that we are using `describe` wrong and it's working by chance, but this behaviour is currently going to make every PR fail, so I've changed the todo to `TODO [>=3.0.0]`, i.e. it's on us to fix before we ship v3.

papb: Hmm, great catch, sorry about that, I am pretty sure the `sync` type signature is the one wrong then. I will look into it later.

PR #228, opened 2020-05-26
ci: create starter github action

mmkal:
This adds a placeholder github action. The file needs to exist on master before the action will run on branches/forks, so I just want to put a v simple one in before iterating on it.

@papb after the test refactor we're at pretty much 100% code coverage. After this, I'd like to add coverage reporting to PRs/branches - jest/github actions/codecov make it really easy to do without having to setup or configure permissions for any additional accounts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,10 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


mmkal:
Looks like I'm wrong, new github actions can run on branches. I'll close this and open a PR with an action which actually does something.

PR #229, opened 2020-05-26
test: code coverage

mmkal:
Add a github action with code coverage reporting.

Also close some of the gaps in code coverage - most were unused code or error conditions.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+        sequelize-version: ['5', 'next']
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: run ci
+      run: |
+        npm i sequelize@$(echo $SEQUELIZE)
```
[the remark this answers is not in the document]
mmkal: 🤦 you are absolutely right.
```suggestion
        npm i sequelize@${{ matrix.sequelize-version }}
```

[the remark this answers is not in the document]
mmkal: That fixed it - should be now at parity with travis ci. As a follow-up, I think this step could become a conditional one - i.e. do nothing if sequelize-version != next. This way we risk testing something slightly different from what we release in future.

test/storage/sequelize.test.ts
```diff
@@ -340,6 +359,16 @@ describe('sequelize', () => {
 	});
 
 	describe('executed', () => {
+		it('validates migration types', async () => {
+			const storage = new Storage({ sequelize: helper.sequelize });
+
+			jest.spyOn(storage.model, 'findAll').mockResolvedValueOnce([{ name: 123 } as any]);
```
papb, line 365: Question: I noticed in the code for the `"lazy-initializes model"` test, there are calls to `.mockClear()` the spies, while this test does not have it. Why is it different here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: run ci
+      run: |
+        npm i sequelize@${{ matrix.sequelize-version }}
+        npm install
+        npm run build
```
papb, line 21: Interesting, `npm run build` is not present in `.travis-ci.yml`... Is it because the `prepare` hook is ran automatically?

test/storage/sequelize.test.ts
mmkal: No good reason. To be hones the `.mockClear()` in the other places doesn't really do anything, since nobody else is using the spy. And even if they were, the usual way to ensure tests don't pollute each other is to call `jest.clearAllMocks()` in a `beforeEach`. So I'll remove the `mockClear`s from the other test.

.github/workflows/ci.yml
mmkal: Yes, good catch, this is building twice unnecessarily. Have removed.

mmkal:
@papb sure, I've deleted travis.yml as part of this PR. I had originally thought we might want both for a while in case something goes wrong with GA, but that'll just be annoying to maintain and if we _really_ need to for some reason, it's easy to bring travis back.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: run ci
+      run: |
+        npm install
+        npm run lint
+        npm run test -- --coverage
+      env:
+        CI: true
+    - name: Coverage
+      uses: codecov/codecov-action@v1
+      if: matrix.node-version == '12.x' && matrix.sequelize-version == null
```
papb, line 28: Questions (I'm learning how to setup GitHub CI by seeing what you wrote):

* Why is there a `-` in some of the lines (12, 13, 17, 19, 26) but not all of them?
* I am confused by indentation, it looks like the `if:` lines are 'inside' the above `name`/`uses` specifications but they are not indented?

(These might be more yaml questions than GitHub CI questions)

mmkal: Yeah, yaml takes a bit of getting used to - it's one of those languages that's easier to write than read. The `-`s are for list items, so each `-` under `steps` is a single step, represented by an object with properties (`uses`, `run`, `with`). An easy way to tell what's going on is with a yaml-to-json converter like https://www.convertjson.com/yaml-to-json.htm - you can paste in [this file](https://raw.githubusercontent.com/sequelize/umzug/5b1bfff23059dbd43522ebceb6e9909f624be91d/.github/workflows/ci.yml)

GitHub actions docs are also pretty good at explaining what each section does.

papb: Nice!! Thanks for that link. I got it now.

P.S. you can merge the PR :)

mmkal: Forgot to actually merge, thanks for the reminder!

Issue #230, opened 2020-05-29
down({ to: undefined }) means revert all migrations?


fwh1990:
`umzug.down({ to: undefined })` has the same behavior with `umzug.down({ to: 0 })`, is the logic correct?

Maybe it should be considered as `umzug.down({})` or `umzug.down()`

mmkal:
I agree, `{ to: undefined }` should be equivalent to `{}`. @papb what do you say we update it to work that way for the v3? This would technically be a breaking change though, so I don't think we can release a fix on v2.

fwh1990:
@mmkal  Fortunately, I'm using v3-beta, thanks for your effort.

papb:
Hello, I agree, `{ to: undefined }` should be equivalent to `{}`

PR #231, opened 2020-06-07
fix: to: undefined shouldn't be like to: 0

mmkal:
fixes https://github.com/sequelize/umzug/issues/230

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -56,6 +56,7 @@ module.exports = {
     '@typescript-eslint/no-unsafe-member-access': 'off',	
     '@typescript-eslint/no-unsafe-call': 'off',	
     '@typescript-eslint/unified-signatures': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
 
     // xo defaults that overlap with prettier
     'comma-dangle': 'off',
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
 	},
 	"scripts": {
 		"build": "del-cli lib && tsc",
-		"lint": "eslint --ext .js,.ts .",
+		"lint": "eslint --ext .js,.ts . --max-warnings 0",
 		"prepare": "npm run build",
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
     '@typescript-eslint/no-unsafe-member-access': 'off',	
     '@typescript-eslint/no-unsafe-call': 'off',	
     '@typescript-eslint/unified-signatures': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
```
mmkal, line 59: I turned this off, since I couldn't think of a scenario where it usefully catches anything. Maybe if this becomes a _really_ active repo one day, and we get a lot of pull requests where people accidentally forget to implement their functions or something?? For now, it's unnecessarily noisy when defining a function like `const noop = () => {}` .

package.json
```diff
@@ -49,7 +49,7 @@
 	},
 	"scripts": {
 		"build": "del-cli lib && tsc",
-		"lint": "eslint --ext .js,.ts .",
+		"lint": "eslint --ext .js,.ts . --max-warnings 0",
```
mmkal, line 52: We are currently at zero warnings. This will help us stay that way!

src/umzug.ts
```diff
@@ -211,7 +211,10 @@ export class Umzug extends EventEmitter {
 			return (await this.executed()).reverse();
 		};
 
-		if (!options || Object.keys(options).length === 0) {
+		// todo [>=3.0.0] restrict what _run receives, this check is brittle
+		// it's ok for this method to have pretty ambiguous inputs since it's public
+		// but _run is private so we should be more precise about what we pass to it.
+		if (!options || Object.keys(options).every(k => typeof options[k] === 'undefined')) {
```
mmkal, line 217: This is the actual fix. Not particularly elegant but does the job.

PR #232, opened 2020-07-05
fix: sequelize latest

mmkal:
`sequelize@next` no longer seems to exist - this switches CI over to `sequelize@latest`.

### DIFF
--- a/test/storage/sequelize.test.ts
+++ b/test/storage/sequelize.test.ts
@@ -10,7 +10,12 @@ import jetpack = require('fs-jetpack');
 
 // TODO [>=3.0.0]: Investigate whether we are mis-using `model.describe()` here, and get rid of `any`.
 // See https://github.com/sequelize/umzug/pull/226 and https://github.com/sequelize/sequelize/issues/12296 for details
-const describeModel = (model: any) => model.describe();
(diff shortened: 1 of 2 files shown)


test/storage/sequelize.test.ts
```diff
@@ -10,7 +10,12 @@ import jetpack = require('fs-jetpack');
 
 // TODO [>=3.0.0]: Investigate whether we are mis-using `model.describe()` here, and get rid of `any`.
 // See https://github.com/sequelize/umzug/pull/226 and https://github.com/sequelize/sequelize/issues/12296 for details
-const describeModel = (model: any) => model.describe();
+const describeModel = (model: any) =>
+	model.describe().then(d => {
+		// FIXME [sequelize@>=6] remove this hack when only sequelize>=6 is supported
+		Object.keys(d).forEach(k => delete d[k].unique);
```
papb, line 16: :thinking: Odd, why is this being done?

mmkal: We're asserting in a few places on the output of this function. The new version of sequelize includes an extra property `unique`. So that tests pass for old and new sequelize, this deletes it.

PR #233, opened 2020-07-05
feat: glob-based constructor

mmkal:
@papb I've had this on the back-burner for a little while but thought I'd open a PR to start getting feedback. This aims to be the first step to simplifying and replacing the many (and sometimes overlapping) constructor options the current incarnation can take. `README.md`, `test/umzug.test.ts` and `test/sequelize.test.ts` have some usage examples, and the readme has some explanations of the "breaking" changes. "breaking" is in quotes because there's nothing removed just yet, it's only an additional function - but if this can go in, we should try out the new API on some known consumers, like @sequelize/cli, and @slonik/migrator, and if we're happy with the changes, the old API should be torn out.


mmkal:
blocked by #232

README.md
```diff
@@ -74,14 +67,56 @@ async function up(queryInterface) {
...
+const sequelize = new Sequelize({ dialect: 'sqlite', storage: './db.sqlite' });
+
+const umzug = getUmzug({
+  migrations: { glob: 'migrations/*.ts' },
+  context: sequelize.getQueryInterface(),
+  storage: new SequelizeStorage({ sequelize }),
+});
+
+// export the type helper exposed by umzug, which will have the `context` variable inferred
+export type Migration = typeof umzug._types.migration;
```
papb, line 96: I think it would be better if we simply provided the Migration type directly to end-users, no? So they can just do `import type { Migration } from 'umzug'` in their migration files, what do you think?

mmkal: We can also do that, but this is a little more specific. `umzug._types.migration` knows about `context` - so in this case the function would have a parameter with a strongly-typed sequelize query interface. If we expose a static interface it won't carry any information about context. But I do accept the `typeof umzug._types.migration` format is awkward. I'll think if there's a better way to get the same result.

README.md
```diff
@@ -230,36 +265,92 @@ module.exports = {
...
-      sequelize.getQueryInterface()
-    ]
-  )
+    })
+  },
+  context: sequelize.getQueryInterface(),
+});
+```
+
+### Upgrading from v2.x
```
papb, line 319: Fantastic! I really like the changes so far. Will take a closer look in the PR tomorrow.

src/umzug.ts
```diff
@@ -414,3 +414,111 @@ export class Umzug extends EventEmitter {
...
+	up: () => Promise<unknown>;
+	down?: () => Promise<unknown>;
+}
+
+export type MigrationList = Array<{ name: string; path?: string; migration: MigrationContainer }>;
+
+export type InputMigrations<T> =
+	| {
+			/**
+			 * A glob string for migration files. Can also be in the format `[path/to/migrations/*.js', {cwd: 'some/base/dir', ignore: '**ignoreme.js' }]`
```
papb, line 431:
```suggestion
			 * A glob string for migration files. Can also be in the format `['path/to/migrations/*.js', {cwd: 'some/base/dir', ignore: '**ignoreme.js' }]`
```

papb: Also, I think we should allow an array of globs as well.

src/migration.ts
```diff
@@ -86,12 +82,8 @@ export class Migration {
 			throw new Error('Could not find migration method: ' + method);
 		}
 
-		const result = this.options.migrations.wrap(fn).apply(migration, args);
-
-		if (!isPromise(result)) {
-			throw new Error(`Migration ${this.file} (or wrapper) didn't return a promise`);
```
papb, line 92: Why was this check removed?

README.md
```diff
@@ -21,38 +21,31 @@ _Note: master represents the next major version of umzug - v3 - which is current
...
-  migrations: {
-    path: './migrations',
-    params: [
-      sequelize.getQueryInterface()
-    ]
-  },
-  storage: new SequelizeStorage({ sequelize })
+const umzug = getUmzug({
+  migrations: { glob: 'migrations/*.js' },
+  context: sequelize.getQueryInterface(),
```
papb, line 33: I would expect `context` to be a plain object containing assorted things, like

```suggestion
  context: {
    queryInterface: sequelize.getQueryInterface(),
  }
```

what do you think?

This would allow:

```ts
const context = {
  queryInterface: sequelize.getQueryInterface()
};

export type Context = typeof context;

const umzug = getUmzug({
  migrations: { glob: 'migrations/*.js' },
  storage: new SequelizeStorage({ sequelize }),
  context
};
```

And in the migration file:

```ts
// migrations/00_initial.ts
import type { Migration } from 'umzug';
import type { Context } from '..';
// types will now be available for `queryInterface`
export const up: Migration<Context> = ({ context: queryInterface }) => queryInterface.createTable(...)
```

(of course we would have to tweak the `Migration` type to receive this generic type.

sushantdhiman: Yeah `context` object makes sense, we can add additional options in future as well

src/migration.ts
mmkal: Turning this around - why did we need it it in the first place? `await 123` and `await Promise.resolve(123)` both resolve to `123` anyway.

To answer your question, though - removing the check allows passing in jest mocks created with `jest.fn()` directly in, without having to mock the implementation when we don't really care about it.

README.md
mmkal: About the context being a plain object - that is already possible. Anything that `context` is set to will be passed as-is to the `context` property of each migration function.

mmkal: Re `export type Context` and making `Migration` generic. I would like to do something like that to. Right now it's a bit tricky because we still have the old `Migration` class, so naming gets a bit confusing. What I think we should do long term is something like

```typescript
import { getUmzug, MigrationOf } from 'umzug'

const umzug = getUmzug({
  migrations: { glob: 'migrations/*.js' },
  storage: new SequelizeStorage({ sequelize }),
})

export type Migration = MigrationOf<typeof umzug>
```

Then in the migration file:

```typescript
import { Migration } from '..'

export const up: Migration = ({ context }) => ...
```

That way the migrations themselves become simpler - they only need to import one thing, they don't need to use generics, and they don't have the possibility of importing `Context` from the wrong place or using `any` instead. Since there are in general far more migration files than umzug-entrypoint files, it feels more correct to have the generics live in the entrypoint.

👆 all that may need to go into a follow-up though. I've deliberately left the `Umzug` class alone as much possible. My plan was to consider the old class/constructor overloads as deprecated, then try out the new API with the sequelize CLI and the slonik migrator I maintain. If it goes well then we can move the logic "up" into the main Umzug class. Until then the `MigrationOf` in the above snippet isn't possible, since the `Umzug` class has no knowledge of `context`.

src/umzug.ts
mmkal: Re an array of globs, I don't think it's necessary. [glob](https://npmjs.com/package/glob) supports multiple patterns via `{folder1/migrations/*.js,folder2/migrations/*.js}`. For very complex setups, users can also use `getMigrations`:

```js
import { getMigrations, getUmzug } from 'umzug'

const m1 = getMigrations({ glob: 'folder1/*.js', ignore: '*commonIgnorePattern*' })
const m2 = getMigrations({ glob: 'folder2/*.js', ignore: ['commonIgnorePattern*', '*folder2OnlyIgnorePattern*] })

const umzug = getUmzug({ migrations: [...m1, ...m2] })
```

src/migration.ts
papb: I think this check was here as a safeguard against people forgetting to return their promises in the migration. But maybe you can convince me this shouldn't be our problem...

mmkal:
Closing in favour of https://github.com/sequelize/umzug/pull/325

PR #239, opened 2020-07-08
chore(deps): bump np version

mmkal:
drops package with security vulnerability (url-regex)

@papb @sushantdhiman this one was previously using a git hash rather than npm version (@papb - why was that?), so renovate won't be able to update it. Doing it manually - after this is in, renovate will handle updates for us.


papb:
Hi @mmkal, sorry I missed this, the reason for the specific github url is because I needed the `--no-2fa` flag that someone made a PR but Sindre hadn't merged yet

PR #321, opened 2020-09-18
Update README.md

luwol03:
hello guys,
it took me a long time to figure out, that I have to install Umzug with the beta tag to use the documentation in the master branch. That's why I want this little description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,10 @@ See [this test](./test/test.ts) for another example of Umzug usage.
 
 #### Installation
 
-Umzug is available on npm:
+The beta version of Umzug is available on npm by specifying the correct tag:
 
 ```bash
-npm install umzug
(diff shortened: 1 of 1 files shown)


mmkal (approved):
Looks good - just a note, the beta version is likely going to change soon, so you may want to stick with v2 for the time being.

luwol03:
What does you mean with going to change soon? Are there an other syntax in beta6?

mmkal:
The usage will be slightly different in v3. Not sure yet which beta revision or timeline, but there will be some breaking changes. No functionality is going away, but some of the options will be simplified/consolidated.

papb:
You can use `v3.0.0-beta.5`, but don't forget that there might be breaking changes between what you see in `v3.0.0-beta.5` and `v3.0.0-beta.6` (and the final `v3`)

PR #323, opened 2020-09-19
ignore null values in sequelize storage

luwol03:
Hello guys,

I like umzug a lot. But I get this error.
image843
After a lot of time I figgerd out, that I can't use a table like that to compress the migrations and seeders into one:
```sql
CREATE TABLE ".migrations" (
	"id"	INTEGER,
	"migrations"	VARCHAR(255),
	"seeders"	VARCHAR(255),
	"createdAt"	DATETIME NOT NULL,
	"updatedAt"	DATETIME NOT NULL,
	PRIMARY KEY("id" AUTOINCREMENT)
);
```
I think it would be a good idea to reduce the array insted of map it and throw an error if it isn't a string.

I'm not so familar with typescript. If there are types missing please add it.

### DIFF
--- a/src/storage/sequelize.ts
+++ b/src/storage/sequelize.ts
@@ -143,14 +143,18 @@ export class SequelizeStorage implements UmzugStorage {
 	async executed(): Promise<string[]> {
 		await this.model.sync();
 		const migrations: any[] = await this.model.findAll({ order: [[this.columnName, 'ASC']] });
-		return migrations.map(migration => {
+		return migrations.reduce((migrations, migration) => {
 			const name = migration[this.columnName];
+			if (name === null) {
+				return migrations;
+			}
 			if (typeof name !== 'string') {
 				throw new TypeError(`Unexpected migration name type: expected string, got ${typeof name}`);
 			}
 
-			return name;
-		});
+			migrations.push(name);
+			return migrations;
+		}, []);
 	}
 
 	// TODO remove this
--- a/src/storage/sequelize.ts
+++ b/src/storage/sequelize.ts
@@ -143,18 +143,16 @@ export class SequelizeStorage implements UmzugStorage {
 	async executed(): Promise<string[]> {
 		await this.model.sync();
 		const migrations: any[] = await this.model.findAll({ order: [[this.columnName, 'ASC']] });
-		return migrations.reduce((migrations, migration) => {
-			const name = migration[this.columnName];
-			if (name === null) {
-				return migrations;
-			}
-			if (typeof name !== 'string') {
-				throw new TypeError(`Unexpected migration name type: expected string, got ${typeof name}`);
-			}
+		return migrations
+			.filter((migration) => migration[this.columnName] !== null)
+			.map(migration => {
+				const name = migration[this.columnName];
+				if (typeof name !== 'string') {
+					throw new TypeError(`Unexpected migration name type: expected string, got ${typeof name}`);
+				}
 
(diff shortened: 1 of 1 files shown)


src/storage/sequelize.ts
```diff
@@ -143,14 +143,18 @@ export class SequelizeStorage implements UmzugStorage {
...
+				return migrations;
+			}
 			if (typeof name !== 'string') {
 				throw new TypeError(`Unexpected migration name type: expected string, got ${typeof name}`);
 			}
 
-			return name;
-		});
+			migrations.push(name);
+			return migrations;
```
mmkal, line 156: `reduce` is usually for avoiding mutating arrays, doesn't make much difference in this case but in general better practice to do `return [... migrations, name]` here.

src/storage/sequelize.ts
```diff
@@ -143,14 +143,18 @@ export class SequelizeStorage implements UmzugStorage {
 	async executed(): Promise<string[]> {
 		await this.model.sync();
 		const migrations: any[] = await this.model.findAll({ order: [[this.columnName, 'ASC']] });
-		return migrations.map(migration => {
+		return migrations.reduce((migrations, migration) => {
```
mmkal, line 146: Let's not shadow the `migrations` variable. In fact, the linter should protect us from this. Will check why it's not soon.

src/storage/sequelize.ts
```diff
@@ -143,14 +143,18 @@ export class SequelizeStorage implements UmzugStorage {
...
+			}
 			if (typeof name !== 'string') {
 				throw new TypeError(`Unexpected migration name type: expected string, got ${typeof name}`);
 			}
 
-			return name;
-		});
+			migrations.push(name);
+			return migrations;
+		}, []);
```
mmkal, line 157: Can we change `[]` to `[] as string []`?

src/storage/sequelize.ts
```diff
@@ -143,14 +143,18 @@ export class SequelizeStorage implements UmzugStorage {
 	async executed(): Promise<string[]> {
 		await this.model.sync();
 		const migrations: any[] = await this.model.findAll({ order: [[this.columnName, 'ASC']] });
-		return migrations.map(migration => {
+		return migrations.reduce((migrations, migration) => {
 			const name = migration[this.columnName];
+			if (name === null) {
+				return migrations;
```
mmkal, line 149: Isn't doing this the equivalent of a using `.filter` before `.map`?

mmkal (changes requested):
Comments on the code left inline.

More generally: does this make sense to do? Do we really want to silently filter out migrations instead of throwing? For sure, something is wrong since you're getting unhandled rejections, but I don't think it's wrong _here_.

luwol03:
Year, you are right with the last points. Filter is a better way to handle it.

luwol03:
Hello, are there any updates, because i am very fustrated, now, if this wont be merged, i have tried to use 2 different table names, but always both **new** `SequelizeStorage` instaces uses the same sql table.

mmkal:
@luwol03 as I mentioned before, I'm not convinced ignoring invalid values is a better idea than throwing. Your new approach of a separate umzug instance for seeders makes sense to me.

About seeders going in the wrong table: it's likely because the sequelize storage only creates a table if the model doesn't already exist. Try replacing `tableName` with `modelName` in both umzug constructors.

luwol03:
Ok, thank you with model name it works great..

PR #349, opened 2020-10-13
fix: import sequelize as a type

### LINKED ISSUE #348: Umzug with mongodb fails to compile with tsc
I use `umzug` to run `mongodb` migrations. I also use typescript in my project.
When I try to compile my project I have next error:

> node_modules/umzug/lib/storage/sequelize.d.ts:3:65 - error TS2307: Cannot find module 'sequelize' or its corresponding type declarations.
> 
> 3 import { Sequelize as SequelizeType, Model as ModelClass } from 'sequelize';

Is it possible to use `umzug` without `sequelize` with typescript. Maybe I'm missing something.

mmkal:
also ignore errors when importing, so users not using sequelize storage don't hit typescript errors

Fixes #348

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,7 +34,7 @@ module.exports = {
 		'jest/expect-expect': ['error', { assertFunctionNames: ['expect', 'expectTypeOf', 'verify'] }],
 
 		'@typescript-eslint/ban-types': 'off',
-		'@typescript-eslint/ban-ts-comment': ['warn', { 'ts-expect-error': false }],
+		'@typescript-eslint/ban-ts-comment': 'off', // covered by prefer-ts-expect-error
 		'@typescript-eslint/prefer-function-type': 'error',
 		'@typescript-eslint/restrict-template-expressions': 'error',
 		'@typescript-eslint/no-shadow': 'error',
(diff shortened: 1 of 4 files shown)


papb (commented):
https://github.com/sequelize/umzug/issues/348#issuecomment-708845118

.eslintrc.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
 		'jest/expect-expect': ['error', { assertFunctionNames: ['expect', 'expectTypeOf', 'verify'] }],
 
 		'@typescript-eslint/ban-types': 'off',
-		'@typescript-eslint/ban-ts-comment': ['warn', { 'ts-expect-error': false }],
+		'@typescript-eslint/ban-ts-comment': 'off', // covered by prefer-ts-expect-error
```
papb, line 37: I think `'@typescript-eslint/ban-ts-comment'` checks more things than just `@ts-ignore` and `@ts-expect-error`, doesn't it?

mmkal: You're right. I've changed this to allow ts-ignore and ts-expect-error only with explanations. Even though ts-ignore is whitelisted, prefer-ts-expect-error will still complain if you try to use it, so to use ts-ignore you need an explanation _and_ a eslint-disable-next-line. Which is a good thing IMO, ts-ignore is dangerous and should only be used rarely!

PR #354, opened 2020-10-14
fix: keep extension in migration name

### LINKED ISSUE #353: [v3 beta6] Migration names missing file extensions
Before the new globbing feature, Umzug stored the migration as the full filename (e.g. `123456-foo.js`) but with the new globbing feature it looks like it's storing the actual name of the migration without the file extension.

See line 371, which strips the file extension for the name
https://github.com/sequelize/umzug/blob/006036ae77ed6fbff9d06e89ad4690ff51a9c2a7/src/umzug.ts#L371-L376

And is then used as the migration name in the "name" column of the migration table
https://github.com/sequelize/umzug/blob/006036ae77ed6fbff9d06e89ad4690ff51a9c2a7/src/umzug.ts#L272

This introduces a major breaking change in existing Umzug installations as the name is different and thus tries to apply every migration again and ends up adding duplicate migrations to the migration table.

| name |
|------|
| 123456-foo |
| 123456-foo.js |

mmkal:
Fixes #353

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -368,7 +368,7 @@ export class Umzug<Ctx> extends EventEmitter {
 			const paths = await globAsync(globString, { ...globOptions, absolute: true });
 			return paths.map(unresolvedPath => {
 				const filepath = path.resolve(unresolvedPath);
-				const name = path.basename(filepath, path.extname(filepath));
+				const name = path.basename(filepath);
 				return {
 					name,
 					path: filepath,
--- a/test/umzug.test.ts
+++ b/test/umzug.test.ts
@@ -36,11 +36,11 @@ describe('basic usage', () => {
 
 		await umzug.up();
 
-		expect(names(await umzug.executed())).toEqual(['m1']);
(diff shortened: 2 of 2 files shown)


src/umzug.ts
```diff
@@ -368,7 +368,7 @@ export class Umzug<Ctx> extends EventEmitter {
 			const paths = await globAsync(globString, { ...globOptions, absolute: true });
 			return paths.map(unresolvedPath => {
 				const filepath = path.resolve(unresolvedPath);
-				const name = path.basename(filepath, path.extname(filepath));
+				const name = path.basename(filepath);
```
mmkal, line 371: my only concern is that I have no idea why I did this is in the first place; it's more effort than doing nothing 🤷‍♂️

papb: Also by the way I slightly prefer `path.parse(filepath).name` instead of `path.basename(filepath, path.extname(filepath))` (it's the same thing)

papb (approved):

PR #355, opened 2020-10-14
Pass name, path, context to up/down functions

### LINKED ISSUE #352: [v3 beta6] Conflicting documentation on context injection
According to the release log for beta6, `context` should be injected on the `up` and `down` functions respectively.

```js
migrations: {
    resolve({ path }) {
      const { up, down } = require(path);

      return {
          up: ({ context }) => up(context),        // <-- context here?
 	  down: ({ context }) => down(context)     // <-- context here?
      }
  }
},
```

However, when looking at the PR (https://github.com/sequelize/umzug/pull/325) I can see that the context should come from the `resolve` function, which I found to work for the beta6 release:

```js
migrations: {
    resolve({ path, context }) {      // <-- Context is actually here
      const { up, down } = require(path);

      return {
          up: () => up(context),
 	  down: () => down(context)
      }
  }
},
```

Just wanted to point out some confusion. 🙂

mmkal:
Closes #352 

To make things as clear as possible, I've replaced all `spy.bind(null, blah)` calls with `async () => spy(blah)`, which makes it more explicit that no parameters are expected. This also adds a test to make sure that both the type system and runtime agree that no parameters are expected.

~This will need to merge from master after #354 goes in as there'll be a "soft" merge conflict in the new test.~

~Note that the default resolver does pass parameters to the up and down functions of file-migrations, though~

This updates the `RunnableMigration` type to receive name, path and context as a params object. You _can_ still write migrations which don't use these parameters, but they'll be passed through to make file-migrations and in-memory-migrations have the same signature and avoid confusion.

### DIFF
--- a/test/umzug.test.ts
+++ b/test/umzug.test.ts
@@ -81,7 +81,7 @@ describe('alternate migration inputs', () => {
 		});
 	});
 
-	test('up and down functions should not take parameters', async () => {
+	test('up and down functions using `resolve` should receive parameters', async () => {
 		const spy = jest.fn();
 
 		const syncer = fsSyncer(path.join(__dirname, 'generated/umzug/parameterless-fns'), {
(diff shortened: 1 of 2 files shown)


papb:
Hmmm, somehow I missed this difference. I don't like the fact that the signatures of the `up` and `down` functions are different when they were created by a custom resolver versus the default. This also means that the "trivial" resolver `x => x` actually breaks it all then?

papb (commented):
Hmmm, somehow I missed this difference. I don't like the fact that the signatures of the `up` and `down` functions are different when they were created by a custom resolver versus the default. This also means that the "trivial" resolver `({ path }) => require(path)` actually doesn't work then?

I think `up` and `down` functions should always receive `{ name, path, context }`. But maybe I'm biased by the fact that I took this for granted subconsiously since the beginning :sweat_smile: is there any reason to not pass them?

Edit: now that I think of it I also don't know why the resolver receives `context`, but I don't mind that, it can't hurt, maybe it would be useful in some situation.

mmkal:
@papb you're, right the user experience works be better if you didn't need to remember the difference between a migration file and a migration object. And I can't think of the top of my head why it _wouldn't_ be possible. I'll try it out in this PR.

test/umzug.test.ts
```diff
@@ -78,6 +81,39 @@ describe('alternate migration inputs', () => {
 		});
 	});
 
+	test('up and down functions should not take parameters', async () => {
```
papb, line 84:
```suggestion
	test('resolved up and down functions should take the usual parameters', async () => {
```

test/umzug.test.ts
```diff
@@ -78,6 +81,39 @@ describe('alternate migration inputs', () => {
...
+					down: async downParams => spy('down', resolveParams.name, downParams.name),
+				}),
+			},
+			logger: undefined,
+		});
+
+		await umzug.up();
+
+		expect(spy).toHaveBeenCalledTimes(1);
+		expect(spy).toHaveBeenNthCalledWith(1, 'up', 'migration1.sql', 'migration1.sql');
```
papb, line 107: Can you also spy and test for `.path` and `.context`, instead of just checking `.name`?

papb (changes requested):
Nice!!

PR #356, opened 2020-10-14
chore(renovate): group dev dependencies

mmkal:
@papb last PR for today, I promise!

This groups all dev dependencies together so we'll get less renovate noise. The tradeoff is, if builds fail, it won't be as immediately obvious which package caused the error. I think worth it since errors for dev dependencies are usually pretty easy to understand.

Once this has gone in and created a dependency dashboard issue, I'd propose we add the `schedule:weekly` config too.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,22 +5,13 @@
     ":semanticCommits"
   ],
   "prConcurrentLimit": 4,
+  "dependencyDashboard": true,
+  "dependencyDashboardAutoclose": true,
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


papb:
> @papb last PR for today, I promise!

:smile: this is not any annoyance, please feel free to make dozens of PRs if you feel like it!

> Once this has gone in and created a dependency dashboard issue, I'd propose we add the schedule:weekly config too.

:thinking: Looks interesting, agreed! Good that I'll learn how that works :smile:

papb (approved):

PR #367, opened 2020-10-30
feat: return migration meta from up/down

mmkal:
v2 returned a list of `name` and `file` values from `up` and `down`, which was useful for inspecting what just happened. This adds the equivalent.


papb:
Nice!

PR #368, opened 2020-10-30
chore: turn on typescript strictNullChecks

mmkal:


papb:
Whoops, thank you. Now that I think of it, when I first converted the repo to TS I decided to do it in two parts, leaving the strict null checks for the second part only, but then I forgot about it completely.

What do you think of enabling the new `--noUncheckedIndexedAccess` option from TS 4.1 as well? I like it

PR #370, opened 2020-10-31
fix: remove sequelize type dependency completely

### LINKED ISSUE #348: Umzug with mongodb fails to compile with tsc

mmkal:
ts-ignore comments are stripped from declaration files: https://github.com/microsoft/TypeScript/issues/38628

It'd be possible to use some kind of codegen to add them back in, but it's easier and more correct to define what we need explicitly from the sequelize type definitions.

Fixes #348 (for real this time)

### DIFF
--- a/src/storage/sequelize.ts
+++ b/src/storage/sequelize.ts
@@ -1,12 +1,34 @@
 import { UmzugStorage } from './contract';
 import { SetRequired } from 'type-fest';
-// eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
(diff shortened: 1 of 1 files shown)


papb:
Excellent!

PR #380, opened 2020-11-15
Completed v2 migration snippet

MichielDeMey:
I believe that for the migration snippet to be 100% complete, we should also pass the name to the return value,
otherwise you might end up with an error:

```
SequelizeValidationError: notNull Violation: SequelizeMeta.name cannot be null
```


mmkal:
Thanks @MichielDeMey!

PR #386, opened 2020-11-18
Support `step` in up and down options

mmkal:
I've been playing around with a CLI, and loosely basing some of the features on [knex-migrate](https://npmjs.com/package/knex-migrate). One thing that library has that umzug doesn't is support for running fewer-than-all up migrations, or more-than-one down migrations without having to specify the name of the target migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,11 @@ It is also possible to pass the name of a migration in order to just run the mig
 await umzug.up({ to: '20141101203500-task' });
 ```
 
+To limit the number of migrations that are run, `step` can be used:
+
+```js
+await umzug.up({ step: 2 })
+```
 
 Running specific migrations while ignoring the right order, can be done like this:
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -194,6 +194,11 @@ It is also possible to pass the name of a migration in order to just run the mig
 await umzug.up({ to: '20141101203500-task' });
 ```
 
+To limit the number of migrations that are run, `step` can be used:
+
+```js
+await umzug.up({ step: 2 })
```
papb, line 200:
```suggestion
// This will run the next two migrations
await umzug.up({ step: 2 })
```

README.md
```diff
@@ -210,6 +215,12 @@ const migration = await umzug.down();
 // reverts the last migration and returns it.
 ```
 
+To revert more than one migration, you can use `step`:
+
+```js
+await umzug.down({ step: 2 });
```
papb, line 221:
```suggestion
// This will revert the last two migrations
await umzug.down({ step: 2 })
```

papb (commented):
Just a suggestion for a little documentation improvement.

PR #388, opened 2020-11-19
fix: require ts optimistically

### LINKED ISSUE #387: No resolver for typescript with ts-jest
Version: 3.0.0-beta.8
Storage: SequelizeStorage
Sequelize: 6.1.1
Node: 15.0.1
Dev setup: Typescript@3.9.6

When I try to run test with jest with preset: ts-jest I've got a error back: 

```
No resolver specified for file /home/cojack/Projects/xxx/xxx-sequelize/src/__tests__/migrations/1589746665497-example.ts. You can use the default resolver with typescript files by adding `ts-node` as a dependency and calling `require('ts-node/register')` at the program entrypoint before running migrations. See docs for guidance on how to write a custom resolver.
```

So I have no clue, how this should working. Any advice? I've tried to add a `require('ts-node/register');` at the top of the entry point, where it was a jest.config.js file, also I tried to run jest with configuration like:

```
node -r ts-node/register ./node_modules/.bin/jest ...
```

without success, also tried to add this require into the test file, also without success, same thing happen.

Also repository doesn't provide any example of tests that works that way, only **README** with hidden section explains how it should works, but it doesn't. 

So my question is: does it works? And if so, how I should configure it to works that way.

Regards.

mmkal:
Fixes #387

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -168,10 +168,10 @@ export class Umzug<Ctx> extends EventEmitter {
 		}
 
 		const ext = path.extname(filepath);
-		const canRequire = ext === '.js' || ext in require.extensions;
+		const canRequire = ext === '.js' || ext === '.ts';
 		const languageSpecificHelp: Record<string, string> = {
 			'.ts':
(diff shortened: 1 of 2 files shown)


papb:
Nice!

PR #393, opened 2020-11-20
Log json-able objects instead of strings

mmkal:
Implements https://github.com/sequelize/umzug/discussions/326#discussioncomment-98776

This updates the expected log function format to accept objects instead of strings. Users can always format to strings if they want, but this allows easier integration with json-based logging system. `console` is still an OK logger to pass in and will print out readable logs. 

CC @stefaanMLB

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -9,14 +9,14 @@ const globAsync = promisify(glob);
 
 export type Promisable<T> = T | PromiseLike<T>;
 
-export type LogFn = (message: string) => void;
+export type LogFn = (message: Record<string, unknown>) => void;
 
 /** Constructor options for the Umzug class */
 export interface UmzugOptions<Ctx = never> {
 	/** The migrations that the Umzug instance should perform */
 	migrations: InputMigrations<Ctx>;
 	/** A logging function. Pass `console` to use stdout, or pass in your own logger. Pass `undefined` explicitly to disable logging. */
-	logger: Record<'info' | 'warn' | 'error', LogFn> | undefined;
+	logger: Record<'info' | 'warn' | 'error' | 'debug', LogFn> | undefined;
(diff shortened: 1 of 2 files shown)


mmkal:
@papb merging because I have a few more things I'd like to work on, so going to merge this, but feel free to comment after-the-fact and I can put any updates in a new PR.

src/umzug.ts
```diff
@@ -327,15 +327,15 @@ export class Umzug<Ctx> extends EventEmitter {
...
+			this.logging({ event: 'reverting', name: m.name });
 			this.emit('reverting', m.name, m);
 
 			await m.down?.({ name: m.name, path: m.path, context: this.options.context });
 
 			await this.storage.unlogMigration(m.name);
 
-			const duration = ((Date.now() - start) / 1000).toFixed(3);
-			this.logging(`== ${m.name}: reverted (${duration}s)\n`);
+			const duration = Number.parseFloat(((Date.now() - start) / 1000).toFixed(3));
```
papb, line 337:
```suggestion
			const duration = (Date.now() - start) / 1000;
```

This makes it equivalent to what was done a few lines above. Also I think dividing an integer by 1000 can never introduce floating point errors, so my suggestion is probably strictly equivalent to the current behavior anyway.

papb (commented):
Maybe we can add a suggestion in the readme of a simple json-based logger that users can use?

src/umzug.ts
mmkal: Ended up doing this in a later PR: https://github.com/sequelize/umzug/pull/397/files#diff-98ea65505aaa84092a04a27b8479d32354660e2d23b914f4e44287f9745a8de6R312

mmkal:
Re json logger - I have a branch with a bunch of practical examples, I'll include a logger one.

PR #394, opened 2020-11-23
Typed async events

mmkal:
This switches from nodejs's `EventEmitter` to [emittery](https://npmjs.com/package/emittery) for a few reasons:

- type-safe events. This will prevent people doing things like `.on('migration', ...)` instead of `.on('migrating', ...)` etc. It also types the event payload - especially important now since the payload is changing slightly
- async events. `this.emit(...)` now returns a promise, which we can await. Which means in a follow-up, we can add `beforeAll` and `afterAll` events, which will enable use cases like locking the migration table. That wouldn't be possible with sync events.
-  apparently, less chance of memory leaks 🤷 

The API is smaller than `EventEmitter` which makes this a breaking change by itself. `emittery` only supports single payloads too, meaning we can no longer `emit('migrating', m.name, m)` - but that was pretty redundant anyway, so it makes sense to update to `MigrationParams` as part of the major (v3) release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,20 @@ The `migrations.resolve` parameter replaces `customResolver`. Explicit support f
 
 The constructor option `logging` is replaced by `logger` to allow for `warn` and `error` messages in future. NodeJS's global `console` object can be passed to this. To disable logging, replace `logging: false` with `logger: undefined`.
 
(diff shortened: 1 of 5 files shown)


mmkal:
@papb keeping the merge train going, but again feel free to comment! 🚂

papb:
Great!

papb:
Why did you decide to move from `^` to `~` in our dependencies?

mmkal:
I think tilde is the default for renovate, although I could be wrong because they call out that in most cases you won't get fixes if you use it: https://docs.renovatebot.com/dependency-pinning/#tilde-vs-caret. Happy to re-visit.

papb:
Yeah, I believe caret is better.

PR #397, opened 2020-11-23
Add beforeAll/afterAll events + file locking

mmkal:
Follow-up to #394 

This adds `beforeAll` and `afterAll` events. They run before the first and after the last migration respectively, for both `up` and `down` calls. This allows for locking functionality as requested in https://github.com/sequelize/umzug/discussions/144. Right now, there's a `FileLocker` class which uses the filesystem to implement a very basic locking mechanism. Adding actual database locking is more complex though, since every database system implements locks differently. I plan to use this to add locking to [@slonik/migrator](https://npmjs.com/package/@slonik/migrator), since that's postgresql-only. It _might_ be possible to add a `SequelizeLocker` class similar to `FileLocker` using [transactions](https://sequelize.org/master/manual/transactions.html), but if that's difficult it could come in another follow-up. @papb have you use sequelize locks before?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -96,6 +96,7 @@ module.exports = {
 		'unicorn/no-fn-reference-in-iterator': 'off',
 		'unicorn/no-null': 'off',
 		'unicorn/prevent-abbreviations': 'off',
+		'unicorn/no-useless-undefined': 'off',
 	},
 	overrides: [
 		{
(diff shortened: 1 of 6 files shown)


papb:
Very nice! Great idea. As for using Sequelize transactions to lock the table, that should work, and shouldn't be too hard. I don't have time to try to do it though... Not now at least.

PR #398, opened 2020-11-24
Pass context to storage methods

mmkal:
This enables setting up clients in beforeAll, and attaching them to context so storages can use them.

It's a non-breaking change, since the context is passed in as an additional parameter. It'd be slightly better design to just pass the `logMigration(params)` instead of `logMigration(name, params)`, since `params` contains `name` anyway. But there are lots of breaking changes in v3 already so I didn't want to commit to yet another. @papb I'd be interested to get your take, though. We could simplify (and make breaking) in a follow up before the stable release.

### DIFF
--- a/src/storage/contract.ts
+++ b/src/storage/contract.ts
@@ -1,18 +1,20 @@
-export interface UmzugStorage {
+import { MigrationParams } from '../umzug';
+
+export interface UmzugStorage<Ctx = unknown> {
 	/**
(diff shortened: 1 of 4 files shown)


papb:
Hi!

> It'd be slightly better design to just pass the `logMigration(params)` instead of `logMigration(name, params)`, since `params` contains `name` anyway.

Agreed. I think we can make this breaking change.

PR #411, opened 2020-12-04
Add examples folder

mmkal:
Added a bunch of usage examples that go further than the readme so users can figure out how umzug can fit into their setup.

@papb sorry to put another big review on your plate! The easiest way to review will be to open the folder in the branch: https://github.com/sequelize/umzug/tree/examples-folder/examples in another tab - everything in there is new so the no real benefits of the PR diff view.


papb:
Awesome!! :tada::tada::tada:

PR #416, opened 2020-12-09
Use verror to wrap migration errors

mmkal:
This makes error messages much more helpful, especially when custom resolvers are used which do things like execute SQL directly.

Instead of `syntax error near ';'` users will see `up migration 2000-01-01T12-00.foobar.sql failed: syntax error near ';'`

[verror](https://npmjs.com/package/verror) is pretty widely used so seems appropriate over rolling our own error-wrapping logic. It hasn't had an update in a while though which isn't ideal. @papb if you're aware of another library that's a little more active and does roughly the same thing, I'm all ears.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,7 @@ All events are type-safe, so IDEs will prevent typos and supply strong types for
 
 ### Errors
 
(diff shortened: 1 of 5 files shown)


papb:
This is super cool! I didn't know `verror` but will probably be using it in the future. Not aware of others.

Nice work!

PR #419, opened 2020-12-14
Create context per run

mmkal:
This adds a "lazy" option to the `context` and `migrations` options, which allows for:

1) more customisation in child classes
2) it's a less dangerous flow if two call sites try to call `up` at the same time and do things like mutating `context`, which in theory is possible now

Breaking change to the `beta` tag: `beforeAll`/`afterAll` are replaced by `beforeCommand`/`afterCommand`, which also run on `executed` and `pending` commands too now. For hooks that should only run before migrations are applied/reverted, use:

```js
umzug.on('beforeCommand', ev => {
  if (ev.command !== 'up' && ev.command !== 'down') {
    return
  }
  // ...whatever you were doing in beforeAll/afterAll before
})
```

Easiest to view with https://github.com/sequelize/umzug/pull/419/files?w=1

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -12,7 +12,7 @@ export type Promisable<T> = T | PromiseLike<T>;
 export type LogFn = (message: Record<string, unknown>) => void;
 
 /** Constructor options for the Umzug class */
-export interface UmzugOptions<Ctx = unknown> {
+export interface UmzugOptions<Ctx extends Record<string, unknown> = Record<string, unknown>> {
 	/** The migrations that the Umzug instance should perform */
 	migrations: InputMigrations<Ctx>;
 	/** A logging function. Pass `console` to use stdout, or pass in your own logger. Pass `undefined` explicitly to disable logging. */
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -12,15 +12,15 @@ export type Promisable<T> = T | PromiseLike<T>;
 export type LogFn = (message: Record<string, unknown>) => void;
 
 /** Constructor options for the Umzug class */
-export interface UmzugOptions<Ctx = unknown> {
+export interface UmzugOptions<Ctx extends {} = Record<string, unknown>> {
```
papb, line 15: Why is it `Ctx extends {} = Record<string, unknown>` here...

src/umzug.ts
```diff
@@ -61,10 +62,10 @@ export class MigrationError extends VError {
 	}
 }
 
-export class Umzug<Ctx = unknown> extends emittery.Typed<UmzugEvents<Ctx>> {
+export class Umzug<Ctx extends object = object> extends emittery.Typed<UmzugEvents<Ctx>> {
```
papb, line 65: ...but `Ctx extends object = object` here?

PR #430, opened 2021-01-09
Add prod tsconfig for lib output

mmkal:
@papb this is a pattern I've found useful in a few projects now.

- Have a `tsconfig.json` which includes _all_ source code but has `noEmit: true` so is only used for checking
- Have a `tsconfig.lib.json` which only includes `src`, and extends `tsconfig.json` with `noEmit: false`. This is used to produce the actual library output.
- `tsc -p tsconfig.json` is part of the _lint_ step. It makes sure all the code in the project is ok. `src`, `test`, `examples` etc.
- `tsc -p tsconfig.lib.json` is the _build_ step.

The benefits are:

- All type errors are caught by `npm run lint`. Here's an example from another PR of a change I had to just remember about since it wasn't flagged by CI (the file is under `examples/*`, which wasn't included in `tsconfig.json`: https://github.com/sequelize/umzug/pull/429/commits/435f968a142ef49a7bc7e3e5a4753a6b91d908b6 
- IDEs pick up the correct settings. By default VSCode (and others, I believe), search for a `tsconfig.json` for settings for typescript hints/red squigglies. One example: this project has `strictNullChecks: true`, but since tsconfig.json was ignoring `test`, VSCode wasn't showing null errors. ts-jest still caught them, but it's better DX if the IDE knows what's going on.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,7 @@
 {
-	"include": [
-		"src"
-	],
 	"compilerOptions": {
+		"noEmit": true,
+		"allowJs": true,
 		"outDir": "lib",
 		"target": "es2018",
 		"module": "commonjs",
@@ -20,5 +19,16 @@
 		"noFallthroughCasesInSwitch": true,
 		"noEmitOnError": true,
 		"skipLibCheck": true
-	}
+	},
+	"include": [
+		"src",
+		"test",
+		"examples",
+		"*.md",
+		"*.js",
+		".*.js"
+	],
+	"exclude": [
+		"test/generated"
+	]
 }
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -20,5 +19,16 @@
...
 		"skipLibCheck": true
-	}
+	},
+	"include": [
+		"src",
+		"test",
+		"examples",
+		"*.md",
+		"*.js",
+		".*.js"
```
mmkal, line 29: Since this config is used for eslint, it needs to include all files that we run the linter on. We're linting README.md, and the various `*.config.js`, `*rc.js` files, so need to add them here.

PR #438, opened 2021-02-06
Support tsconfigs with esModuleInterop

### LINKED ISSUE #434: (beta) Build Error: Type 'typeof VError' is not a constructor function type.
- Umzug Version: **3.0.0-beta.14**
- Node Version: `v14.15.4`

Just started using umzug and get this error when i try to build:

```shell
node_modules/umzug/lib/umzug.d.ts:11:45 - error TS2507: Type 'typeof VError' is not a constructor function type.

11 export declare class MigrationError extends VError {
```

mmkal:
Fixes #434 

@papb I remember you suggested using the `import foo = require('foo')` syntax all along!

I'm still not sure this is the ideal solution, but it works and it's clear how it corresponds to `const foo = require('foo')` in the compiled js.

CommonJS vs ES modules is still v confusing to me, but probably setting `"type": "module"` in package.json would make sense at some point: https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1

### DIFF
--- a/examples/3.raw-sql/umzug.ts
+++ b/examples/3.raw-sql/umzug.ts
@@ -1,7 +1,7 @@
 import { Umzug } from 'umzug';
 import { Sequelize } from 'sequelize';
-import * as path from 'path';
-import * as fs from 'fs';
+import path = require('path');
+import fs = require('fs');
 
 const getRawSqlClient = () => {
 	// this implementation happens to use sequelize, but you may want to use a specialised sql client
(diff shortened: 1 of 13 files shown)


papb:
Yeah, haha! Although I didn't know it had anything to do with that tsconfig flag. I was just following Sindre's decision after he had a lot of headache with that.

PR #453, opened 2021-04-06
Add ability to retrieve context asynchronously before the migrations run

alefi:
It often needs that we should do some preparation prior to context could be provided to migrations modules. It could be the initialization of database backend storages, etc. The old behavior keeps untouched, but this PR adds the advantage to use the asynchronous functions to retrieve context.

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -228,10 +228,7 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 	}
 
 	protected async runCommand<T>(command: string, cb: (commandParams: { context: Ctx }) => Promise<T>): Promise<T> {
-		const context: Ctx =
-			typeof this.options.context === 'function'
-				? (this.options.context as () => Ctx)()
-				: ((this.options.context ?? {}) as Ctx);
+		const context = await this.getContext();
 
 		await this.emit('beforeCommand', { command, context });
 		try {
@@ -481,6 +478,30 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 
 			return migration;
 		});
+	}
+
+	private async getContext(): Promise<Ctx> {
+		const isAsyncFunc = (_ctx: object | Function) => {
+			const _ctxProto = Object.prototype.toString.call(_ctx);
+			return _ctxProto === '[object AsyncFunction]';
+		};
+
+		const isPromise = (_ctx: object | Function) => {
+			const _ctxProto = Object.prototype.toString.call(_ctx);
+			return _ctxProto === '[object Promise]';
+		};
+
+		let { context = {} as Ctx }: UmzugOptions<Ctx> = this.options;
+
+		if (isAsyncFunc(context)) {
+			context = await (context as () => Promise<Ctx>)();
+		} else if (isPromise(context)) {
+			context = await Promise.resolve(context as Promise<Ctx>);
+		} else if (typeof context === 'function') {
+			context = (context as () => Ctx)();
+		}
+
+		return context;
 	}
 
 	/** helper for parsing input migrations into a callback returning a list of ready-to-run migrations */
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -481,11 +481,6 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 	}
 
 	private async getContext(): Promise<Ctx> {
-		const isAsyncFunc = (_ctx: object | Function) => {
-			const _ctxProto = Object.prototype.toString.call(_ctx);
-			return _ctxProto === '[object AsyncFunction]';
-		};
-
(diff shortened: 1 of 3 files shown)


alefi:
Hi,
Could someone clarify what approximate time needed to approve or reject this PR? I opened it about 2 weeks ago, and there are no reactions till now. Thanks.

src/umzug.ts
```diff
@@ -483,6 +480,30 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 		});
 	}
 
+	private async getContext(): Promise<Ctx> {
+		const isAsyncFunc = (_ctx: object | Function) => {
```
mmkal, line 484: This seems brittle and not necessary. If it's a sync or async function we can do `await fn()` to gets the result

alefi: You're right, thanks. I've got checked it out. Please review.

alefi:
Hi, is there any chance that this PR would merge in near future? Thanks.

src/umzug.ts
```diff
@@ -483,6 +480,23 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 		});
 	}
 
+	private async getContext(): Promise<Ctx> {
+		const isPromise = (_ctx: object | Function) => {
```
mmkal, line 484: Do we need this? Can't we just check if it's a function and `await context()` if it is?

mmkal (commented):
Sorry for the slow response here. I am 👍 on the idea but I don't think we should be checking if the return value of the context-getter is a function.

src/umzug.ts
alefi: Unfortunately, Promise isn't a function and this comparison is required. We could use different mechanics to detect promise, such either `typeof _ctx.constructor === 'function';` or `Boolean(typeof value.then === 'function');`, but in my opinion, current implementation Is better way to do that. WDYT?

mmkal: I mostly mean that you can await a non-promise too. So we can keep lines 231-234 basically the same but put an `await` before each ternary statement option, then I don't think we'd even need this function.

alefi: The context retrieving action is a significant one, and grouping it into the method `getContext` increases code readability. As for rest, if we would add `await` before each ternary, the Promise branch won't execute at all.

mmkal: That's true. I'm fine with keepinh getContext, encapsulating the logic in a function, as long as we get rid of this `toString` based promise checking. It's brittle (likely won't work with non-native promises like bluebird, etc.) and it doesn't seem necessary.

alefi: I changed the Promise detection implementation. Please review.

alefi: Any updates on it? :)

src/umzug.ts
```diff
@@ -483,6 +480,22 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
...
+
+		let { context = {} as Ctx }: UmzugOptions<Ctx> = this.options;
+
+		if (isPromise(context)) {
+			context = await Promise.resolve(context as Promise<Ctx>);
+		} else if (typeof context === 'function') {
+			context = await (context as () => Promise<Ctx> | Ctx)();
+		}
+
+		return context;
```
mmkal, line 496: Sorry, I clearly haven't been explaining this very well. We do not need to check if the context is a promise. The fact that this function is `async` will ensure it's one anyway, and a the js engine will flatten promises for us. The implementation of this method can and should be much simpler and have less risk of gotchas.

```suggestion
		const { context = {} } = this.options;
		return typeof context === 'function' ? context() : context;
```

The tests you added are good - they should pass if you make this change. Assuming they do, I'm happy to merge.

test/umzug.test.ts
```diff
@@ -91,6 +91,49 @@ describe('custom context', () => {
...
+		const getContext = async () => {
+			// It guarantees the initialization scripts or asynchronous stuff finished their work
+			// before the actual migrations workflow begins.
+			// Eg: const externalData = await retrieveExternalData();
+			await sleep(100);
+
+			return {
+				// Eg: externalData,
+				innerValue: 'text',
+			};
```
mmkal, line 106: nit: this comment is confusing/not necessary - the one above explaining the use case you're trying to replicate is, but no need for the extra comment here:
```suggestion
			return { innerValue: 'text' };
```

mmkal (changes requested):

src/umzug.ts
mmkal: I left a suggestion to explain what I mean. Please add a failing test if you think my suggestion doesn't cover everything.

alefi: I simplified the method according to your suggestion. Thanks!

mmkal (approved):

mmkal:
Released as 3.0.0-beta.16

PR #464, opened 2021-05-17
Add context to migrator._types

adrienduchemin:


mmkal (changes requested):
I think there's value to this, but just FYI you can achieve this via inference:

```ts
type Context = typeof umzug extends Umzug<infer Ctx> ? Ctx : never
```

Assuming you're aware of that, could you add a test, similar to this one: https://github.com/sequelize/umzug/blob/4bf38052cfd40acbf10263c5f0f0f8b52ad5e519/test/umzug.test.ts#L784-L798

Can be very simple, something like

```ts
const umzug = new Umzug({
  migrations: { glob: '*/*.ts' },
  context: { someCustomSqlClient: {} },
  logger: undefined,
});

expectTypeOf(umzug._types).toMatchTypeOf<{ context: { someCustomSqlClient: {} } }>();
```

adrienduchemin:
Thanks for the tip, I was not aware ! I find this is a cleaner version of achieving it though.
If you still think there's also value to it, I added the requested test.

mmkal (approved):

PR #493, opened 2021-08-27
Add documentation for configuring Umzug migration parameters

### LINKED ISSUE #488: Supported sequelize and sequelize-cli versions should be documented prominently
When using the following library versions (all latest available at time of writing)
- sequilize: 6.6.5
- sequelize-cli: 6.2.0
- umzug: 3.0.0-beta.16

After following the [sequelize-cli documentation](https://sequelize.org/master/manual/migrations.html) and running the following commands to create and validate up and down behaviors for a simple migration
`npx sequelize-cli model:generate --name User --attributes firstName:string,lastName:string,email:string`
`npx sequelize-cli db:migrate`
`npx sequelize-cli db:migrate:undo`

Migration created by sequelize-cli:
```
'use strict'
module.exports = {
    up: async (queryInterface, Sequelize) => {
        await queryInterface.createTable('Users', {
            id: {
                allowNull: false,
                autoIncrement: true,
                primaryKey: true,
                type: Sequelize.INTEGER,
            },
            firstName: {
                type: Sequelize.STRING,
            },
            lastName: {
                type: Sequelize.STRING,
            },
            email: {
                type: Sequelize.STRING,
            },
            createdAt: {
                allowNull: false,
                type: Sequelize.DATE,
            },
            updatedAt: {
                allowNull: false,
                type: Sequelize.DATE,
            },
        })
    },
    down: async (queryInterface, Sequelize) => {
        await queryInterface.dropTable('Users')
    },
}
```

When you setup Umzug with the following configuration defined in the Reame
```    
const umzug = new Umzug({
    migrations: { glob: 'migrations/*.js' },
    storage: new SequelizeStorage({ sequelize }),
    logger: console,
});
await umzug.up()
```

You get the following error (slightly modified for readability) when the migration (which was found successfully) is attempted to be applied by Umzug
```
{ 
event: 'migrating', name: '20210810150559-create-user.js'
Migration 20210810150559-create-user.js (up) failed: 
Cannot read property 'INTEGER' of undefined
}
```

This is because the migration at the top of this issue, which was generated by sequilize-cli, requires the following to parameters to be passed to the `up` and `down` methods
- `queryInterface`
- `Sequelize`

From what I can tell in the documentation it does not look like there is a way in Umzug@3 to pass this additional parameter (it seems Umzug@2 has support for a `params` option in config) which means (I believe) that Umzug does not currently support the latest version of sequilize / sequilize-cli. 

If there is a specific version of these library that the latest version of Umzug supports up to, they should be added to the Readme prominently so new users know what version of libraries to use to generate migrations in a way that Umzug can digest and run correctly.

trentprynn:
Resolves #488

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,40 @@ const umzug = new Umzug({
 });
 ```
 
+#### Modifying the parameters passed to your migration methods
+
+Sometimes it's necessary to modify the parameters `umzug` will pass to your migration methods when the library calls the `up` and `down` methods for each migration. This is the case when using migrations currently generated using `sequilize-cli`. In this case you can use the `resolve` fuction during migration configuration to determine which parameters will be passed to the relevant method
+
+```js
+import { Sequelize } from 'sequelize'
+import { Umzug, SequelizeStorage } from 'umzug'
+
+const sequelize = new Sequelize(
+    ...
+)
+
+const umzug = new Umzug({
+    migrations: {
+        glob: 'migrations/*.js',
+        resolve: ({ name, path, context }) => {
+            const migration = require(path)
+            return {
+                // adjust the parameters Umzug will
+                // pass to migration methods when called
+                name,
+                up: async () => migration.up(context, Sequelize),
+                down: async () => migration.down(context, Sequelize),
+            }
+        },
+    },
+    context: sequelize.getQueryInterface(),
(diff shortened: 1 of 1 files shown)


mmkal:
Thanks @trentprynn

Issue #540, opened 2022-02-02
Umzug UP not working with aws lambda function handler. [programmaticaly]


srikantmatihali:
When trying to do migrations from nestApp + Sequelize ORM + docker database, using AWS sam invoke method from local, up migrations are not working. I have initialized the umzup.up() trigger from lambdaenty.ts as below
```
     async function bootstrap() {
             const expressServer = express();
             const nestApp = await NestFactory.create(
                             AppModule,
                             new ExpressAdapter(expressServer),
                            );
              nestApp.use(eventContext());
              sharedBootstrap(nestApp);
              await nestApp.init();
              return serverlessExpress.createServer(expressServer);
          }
          export const handler = (event: any, context: Context) => {
              if (!lambdaProxy) {
                 bootstrap().then((server) => {
                 lambdaProxy = server;
                 serverlessExpress.proxy(lambdaProxy, event, context);
                 (async () => {
                   try {
                  console.log('MIGRATOR::');
                  const umzug = new Umzug({
                    migrations: { glob: 'src/migrations/*.ts' },
                    context: sequelize.getQueryInterface(),
                    storage: new SequelizeStorage({ sequelize }),
                    logger: console,
                  });
                  await umzug
                    .pending()
                    .then((migrations: any) => {
                      console.log('pending ?  : ', JSON.stringify(migrations));
          
                      async () => {
                        //BELOW FUNCTION IS TRIGGERING BUT NOT GETTING MIGRATION LOADED.
                        await umzug.up();
                      };
                    })
                    .catch((e: any) => console.log('error2 ? ', e));
                } finally {
                  await sequelize.connectionManager.close();
                }
              })();
            });
            } else {
            serverlessExpress.proxy(lambdaProxy, event, context);
           }
          }; 
```
Has anyone faced a similar issue before or any suggestion to handle programmatic up of migration files in aws function handler trigger.

stack overflow question on same: https://stackoverflow.com/questions/70951147/umzug-migration-up-with-aws-lambda-function-not-working.

srikantmatihali:
I am able to solve the issue after lot of tries.  I seperated out the sequelize connection code and called it from app side and triggered from lambdaentry

**lambdaEntry.js file.**
`async function bootstrap(uuid = null) {
  console.log('Calling bootstrap');
  const expressServer = express();
  const nestApp = await NestFactory.create(
    AppModule,
    new ExpressAdapter(expressServer),
  );
  nestApp.use(eventContext());
[...]

PR #543, opened 2022-03-13
Aws lamda trigger srikanth

srikantmatihali:


mmkal:
Thanks for taking the time to do this. I don’t think this is something it makes sense to keep in the umzug repo itself. If you can refactor to make it much more minimal (maybe, a lambda which _only_ has dependencies of, say, `sequelize` and `@types/aws-lambda`). All this nestjs, express, aws-serverless-express boilerplate isn’t relevant to this package (and I don’t want to learn nest just to maintain this example). Otherwise, it’s better as a standalone repo.

srikantmatihali:
> Thanks for taking the time to do this. I don’t think this is something it makes sense to keep in the umzug repo itself. If you can refactor to make it much more minimal (maybe, a lambda which _only_ has dependencies of, say, `sequelize` and `@types/aws-lambda`). All this nestjs, express, aws-serverless-express boilerplate isn’t relevant to this package (and I don’t want to learn nest just to maintain this example). Otherwise, it’s better as a standalone repo.

Thank you, https://github.com/sequelize/umzug/issues/540 is the issue I was facing, when trying to do the same as you suggest. I never got that working, hence did work around and got this up. I will have this seperate repo if you insist. thanks for feedback.

PR #565, opened 2022-08-13
fix: verify pending correctly for custom templates

### LINKED ISSUE #559: CLI create: "Expected to be a pending migration but it wasn't!" when using create.template name transformation
Looks like Umzug isn't handling the renaming of the inbound filename terribly well. I could make the "error" go away by adding the flag it mentions, but that just masks the problem.

## Environment

Umzug 3.1.1
SequelizeStorage
Node 16 & 18
TypeScript, CommonJS

## Relevant code snippet

```typescript
	return new Umzug<QueryInterface>({
		context: sequelize?.getQueryInterface(),
		create: {
			folder: migrationsDir,
			template: (filepath): [string, string][] => {
				return [
					[
						`${filepath.toLowerCase()}.m.ts`,
						`import { DataTypes, QueryInterface } from "sequelize";
```

## Example execution
```sh
$ ts-node --files -- src/cli "create" "--name" "asdfQWER"

{
  event: 'created',
  path: 'REDACTED/src/migrations/2022.08.05t21.35.28.asdfqwer.m.ts'
}

Error: Expected REDACTED/src/migrations/2022.08.05T21.35.28.asdfQWER to be a pending migration but it wasn't! You should investigate this. Use --skip-verify to bypass this error.
```

The file gets created, so this is more of a warning than an actual error, but it does cause confusion for the team members.

mmkal:
Fixes #559 

CC @kf6kjg

### DIFF
--- a/test/umzug.test.ts
+++ b/test/umzug.test.ts
@@ -93,6 +93,27 @@ describe('custom context', () => {
 		expect(spy).toHaveBeenCalledTimes(1);
 	});
 
+	test(`create with custom template extension doesn't cause bogus warning`, async () => {
+		const syncer = fsSyncer(path.join(__dirname, 'generated/create-custom-template'), {});
+		syncer.sync();
+
+		const umzug = new Umzug({
+			migrations: {
+				glob: ['*.js', { cwd: syncer.baseDir }],
+			},
+			logger: undefined,
+			create: {
+				folder: syncer.baseDir,
+				template: filepath => [[`${filepath}.x.js`, `/* custom template */`]],
+			},
+		});
+
+		await umzug.create({ name: 'test' });
+		const pending = names(await umzug.pending());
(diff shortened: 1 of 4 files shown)


test/umzug.test.ts
```diff
@@ -93,6 +93,27 @@ describe('custom context', () => {
...
+				glob: ['*.js', { cwd: syncer.baseDir }],
+			},
+			logger: undefined,
+			create: {
+				folder: syncer.baseDir,
+				template: filepath => [[`${filepath}.x.js`, `/* custom template */`]],
+			},
+		});
+
+		await umzug.create({ name: 'test' });
```
mmkal, line 111: Should have pushed the test in a separate commit to prove this but: this was reproducing the bug described in #559 until the change to `src/umzug.ts` above.

PR #575, opened 2022-09-04
Fix false positive Confusing Ordering warning

### LINKED ISSUE #574: False positive Confusing Ordering warning
Hello. I'm trying to setup Umzug with a migrations file structure like this:

```
migrations/
  0001_somename/
    migration.sql
  0002_somename/
    migration.sql
```

I assume I should have no issues with the ordering based on this description:

https://github.com/sequelize/umzug/blob/91d6a4b197747d4d19c45811a597fbefc9e39851/README.md?plain=1#L394-L396

But I have an issue with the `.create()` method. It produces a "confusing ordering" error. And I think there's a bug because the following code basically compares `"migration.sql"` to a new name, rather than comparing full paths of migrations:

https://github.com/sequelize/umzug/blob/91d6a4b197747d4d19c45811a597fbefc9e39851/src/umzug.ts#L358

---

> the full umzug version you're using (e.g. v2.3.0 (latest stable), or v3.x (beta version)

umzug v3.2.1

> which storage you're using - e.g. SequelizeStorage, JsonStorage, a custom one, etc.

A custom storage

> relevant surrounding information like sequelize version if applicable

n/a

> node version, if it might be relevant

v16.16.0

> dev setup. Are you using typescript? Vanilla commonjs?

TypeScript

> repro steps with as much detail as possible

Please let me know if you need this, but I hope it should be clear from the description

rpominov:
Fixes #574 .

I've also noticed a vulnerabilities warning from NPM:

```
$ npm i

> umzug@3.2.1 prepare
> npm run build


> umzug@3.2.1 build
> del-cli lib && tsc -p tsconfig.lib.json


added 1031 packages, and audited 1032 packages in 40s

147 packages are looking for funding
  run `npm fund` for details

7 vulnerabilities (6 moderate, 1 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
```

I've run `npm audit fix` which updated some packages in `package-lock.json` and fixed the `1 high` vulnerability, but `6 moderate` still remain:

```
$ npm audit fix

changed 3 packages, and audited 1032 packages in 1s

147 packages are looking for funding
  run `npm fund` for details

# npm audit report

got  <11.8.5
Severity: moderate
Got allows a redirect to a UNIX socket - https://github.com/advisories/GHSA-pfrx-2q88-qq97
fix available via `npm audit fix --force`
Will install np@3.0.4, which is a breaking change
node_modules/got
node_modules/package-json/node_modules/got
  npm-name  <=6.0.1
  Depends on vulnerable versions of got
  node_modules/npm-name
    np  >=2.2.0
    Depends on vulnerable versions of npm-name
    Depends on vulnerable versions of update-notifier
    node_modules/np
  package-json  <=6.5.0
  Depends on vulnerable versions of got
  node_modules/package-json
    latest-version  0.2.0 - 5.1.0
    Depends on vulnerable versions of package-json
    node_modules/latest-version
      update-notifier  0.2.0 - 5.1.0
      Depends on vulnerable versions of latest-version
      node_modules/update-notifier

6 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force
```

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -355,13 +355,6 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 			const existing = await this.migrations(context);
 			const last = existing[existing.length - 1];
 
-			const confusinglyOrdered = existing.find(e => e.path && path.basename(e.path) > fileBasename);
-			if (confusinglyOrdered && !options.allowConfusingOrdering) {
-				throw new Error(
-					`Can't create ${fileBasename}, since it's unclear if it should run before or after existing migration ${confusinglyOrdered.name}. Use allowConfusingOrdering to bypass this error.`
-				);
-			}
-
 			const folder = options.folder || this.options.create?.folder || (last?.path && path.dirname(last.path));
 
 			if (!folder) {
@@ -375,6 +368,18 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 			const toWrite = template(filepath);
 			if (toWrite.length === 0) {
 				toWrite.push([filepath, '']);
+			}
+
+			if (!options.allowConfusingOrdering) {
+				const confusinglyOrdered = existing.find(e => {
+					const existingPath = e.path;
+					return existingPath && toWrite.some(([newPath]) => existingPath > newPath);
+				});
+				if (confusinglyOrdered) {
+					throw new Error(
+						`Can't create ${fileBasename}, since it's unclear if it should run before or after existing migration ${confusinglyOrdered.name}. Use allowConfusingOrdering to bypass this error.`
+					);
+				}
 			}
 
 			toWrite.forEach(pair => {
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -363,23 +363,20 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 
 			const filepath = path.join(folder, fileBasename);
 
-			const template = this.options.create?.template ?? Umzug.defaultCreationTemplate;
-
-			const toWrite = template(filepath);
-			if (toWrite.length === 0) {
-				toWrite.push([filepath, '']);
-			}
-
 			if (!options.allowConfusingOrdering) {
-				const confusinglyOrdered = existing.find(e => {
-					const existingPath = e.path;
-					return existingPath && toWrite.some(([newPath]) => existingPath >= newPath);
-				});
+				const confusinglyOrdered = existing.find(e => e.path && e.path >= filepath);
 				if (confusinglyOrdered) {
 					throw new Error(
 						`Can't create ${fileBasename}, since it's unclear if it should run before or after existing migration ${confusinglyOrdered.name}. Use allowConfusingOrdering to bypass this error.`
 					);
 				}
+			}
+
+			const template = this.options.create?.template ?? Umzug.defaultCreationTemplate;
+
+			const toWrite = template(filepath);
+			if (toWrite.length === 0) {
+				toWrite.push([filepath, '']);
 			}
 
 			toWrite.forEach(pair => {
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -373,7 +373,7 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 			if (!options.allowConfusingOrdering) {
 				const confusinglyOrdered = existing.find(e => {
 					const existingPath = e.path;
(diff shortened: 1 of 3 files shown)


src/umzug.ts
```diff
@@ -377,6 +370,18 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
...
 			}
 
+			if (!options.allowConfusingOrdering) {
+				const confusinglyOrdered = existing.find(e => {
+					const existingPath = e.path;
+					return existingPath && toWrite.some(([newPath]) => existingPath > newPath);
+				});
+				if (confusinglyOrdered) {
+					throw new Error(
+						`Can't create ${fileBasename}, since it's unclear if it should run before or after existing migration ${confusinglyOrdered.name}. Use allowConfusingOrdering to bypass this error.`
```
rpominov, line 380: There's another inconsistency. The error say "before or after", but the code checks that the new migration comes strictly after all existing migrations.

test/umzug.test.ts
```diff
@@ -114,6 +114,33 @@ describe('custom context', () => {
...
+				template: filepath => [[`${filepath}/migration.js`, `/* custom template */`]],
+			},
+		});
+
+		await umzug.create({ name: 'test1' });
+		await umzug.create({ name: 'test2' });
+		const pending = names(await umzug.pending());
+		expect(pending).toHaveLength(2);
+		expect(pending[0]).toContain('test1');
+		expect(pending[1]).toContain('test2');
```
mmkal, line 141: Nit, can be simplified to:
```suggestion
		expect(pending).toEqual(['test1', 'test2']);
```

src/umzug.ts
```diff
@@ -377,6 +370,18 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
 				toWrite.push([filepath, '']);
 			}
 
+			if (!options.allowConfusingOrdering) {
+				const confusinglyOrdered = existing.find(e => {
+					const existingPath = e.path;
+					return existingPath && toWrite.some(([newPath]) => existingPath >= newPath);
+				});
```
mmkal, line 377: I think this should only compare against `filepath`, not the whole `toWrite` array. `toWrite` tends to be ancillary files after the first entry, like down migrations, which don't necessarily get picked up by `glob`. It might just be luck that the test cases don't trip up on this.

```suggestion
				const confusinglyOrdered = existing.find(e => e.path && filepath >= e.path);
```

mmkal (changes requested):
Thanks for the PR! I put a couple of change requests inline.

test/umzug.test.ts
rpominov: `pending[0]` looks something like this: `"2022.09.04T11.37.50.test1"`.
Should I do `expect(pending).toEqual(['2022.09.04T11.37.50.test1'...`?

test/umzug.test.ts
```diff
@@ -114,6 +114,33 @@ describe('custom context', () => {
...
+	test(`create doesn't cause "confusing oredering" warning when migrations are nested in folders`, async () => {
+		const syncer = fsSyncer(path.join(__dirname, 'generated/create-nested-folders'), {});
+		syncer.sync();
+
+		const umzug = new Umzug({
+			migrations: {
+				glob: ['*/*.js', { cwd: syncer.baseDir }],
+				resolve(params) {
+					const name = path.basename(path.dirname(params.path!));
+					return { name, async up() {} };
```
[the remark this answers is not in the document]
rpominov: Not sure whether by accident but the test fails against main even without `path`.

Added `path` and checked again that it still fails against main.

src/umzug.ts
rpominov: Ah, makes sense. What if the message would explain it more or less how you just explained it? Or would it be to verbose?

`Can't create ${fileBasename}, since it's unclear if it should run before or after existing migration ${confusinglyOrdered.name}. Chronologically it's being created after ${confusinglyOrdered.name}, but alphabetically it comes before. Use allowConfusingOrdering to bypass this error.`

test/umzug.test.ts
mmkal: Oh right. No never mind. Ignore this.

mmkal (approved):

rpominov:
Just noticed that you've approved but did not merge. Please let me know if you need anything from me. No rush, only want to make sure I don't block.

As for the error message, changing it is just a suggestion, I don't have a strong opinion.

PR #585, opened 2022-10-13
Define custom prefixes for creating new migrations

oetherington:
This PR allows the user to specify a custom format for the filename prefix used when creating new migrations. This is currently only possible by returning a different filename from the `template` callback, but this doesn't happen until after ordering is resolved and so can easily lead to "confusing ordering" issues. This is primarily motivated by wanting to avoid dots in the filename timestamp, but could be useful for other reasons too such as using a counter instead of a timestamp.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -11,6 +11,8 @@ export type Promisable<T> = T | PromiseLike<T>;
 
 export type LogFn = (message: Record<string, unknown>) => void;
 
+export type MigrationPrefix = 'TIMESTAMP' | 'DATE' | 'NONE';
+
 /** Constructor options for the Umzug class */
 export type UmzugOptions<Ctx extends {} = Record<string, unknown>> = {
 	/** The migrations that the Umzug instance should perform */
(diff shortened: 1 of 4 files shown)


src/umzug.ts
```diff
@@ -332,23 +333,28 @@ export class Umzug<Ctx extends object = object> extends emittery<UmzugEvents<Ctx
...
+			NONE: '',
+		};
+		return prefixes[prefix];
+	}
+
 	async create(options: {
 		name: string;
 		folder?: string;
-		prefix?: 'TIMESTAMP' | 'DATE' | 'NONE';
+		prefix?: MigrationPrefix;
```
mmkal, line 349: It's a bit confusing having two places to setup the prefix. Can we deprecate this one with a warning message?

src/types.ts
```diff
@@ -34,6 +36,11 @@ export type UmzugOptions<Ctx extends {} = Record<string, unknown>> = {
 		 * in the same folder as the last existing migration. The value here can be overriden by passing `folder` when calling `create`.
 		 */
 		folder?: string;
+		/**
+		 * A function for generating custom prefixes for migration files when using `create`. If this is not specified the default date formats will
+		 * be used ("1970.01.01T00.00.00" for TIMESTAMP, "1970.01.01" for DATE and "" for NONE)
+		 */
+		prefix?: (prefix: MigrationPrefix) => string;
```
mmkal, line 43:
```suggestion
		prefix?: (params: {name: string}) => string;
```

I like this approach but we can go further. The `MigrationPrefix` type isn't really useful anymore. Let's deprecate it and just pass the name in and let the user define a function that does whatever they need. The default can still be the dot-separated date format that's currently used.

mmkal (commented):
Thurs

Issue #589, opened 2022-12-27
create options template with async method 


ccamensuli:
Proposal for   apply async method on template options create

 use version    "3.2.1"

##  Current
``` js
const umzug = new Umzug({
  migrations: ...,
	create: {
		template: filepath => [
			[filepath, fs.readFileSync('path/to/your/template/file').toString()],
		]
	}
})
```

##  Proposal 
``` js
const umzug = new Umzug({
  migrations: ...,
   create: {
	template: async (filepath) => {
              return [
                [filepath,  await this.renderAsync('path/to/your/template/file')]
              ]
          }
   }
})
```

##    src/umzug.ts   
``` ts
 //  maybe  just add await  when call template method 
 const toWrite = await template(filepath);
 if (toWrite.length === 0) {
      toWrite.push([filepath, '']);
  }
```

mmkal:
@ccamensuli  I'd be happy to accept a PR for this. As long as it's backwards compatible - that is, the types should still allow for a synchronous function.

ccamensuli:
@mmkal   I'm a little short of time to implement this feature at the moment, but when I have it I'll be happy to offer you a PR

thank you for the interest

PR #591, opened 2023-01-12
feat(create-options):#589 create options template with async method

ccamensuli:


src/types.ts
```diff
@@ -28,7 +28,7 @@ export type UmzugOptions<Ctx extends {} = Record<string, unknown>> = {
 		 * Should return an array of [filepath, content] pairs. Usually, only one pair is needed, but to put `down` migrations in a separate
 		 * file, more than one can be returned.
 		 */
-		template?: (filepath: string) => Array<[string, string]>;
+		template?: (filepath: string) => Array<[string, string]> | Promise<Array<[string, string]>>;
```
mmkal, line 31: We have a helper for this

```suggestion
		template?: (filepath: string) => Promisable<Array<[string, string]>>;
```

test/umzug.test.ts
```diff
@@ -114,6 +114,34 @@ describe('custom context', () => {
...
 
+	test(`create with custom template async method`, async () => {
+		const syncer = fsSyncer(path.join(__dirname, 'generated/create-custom-template-async'), {});
+		syncer.sync();
+
+		const methodAsync:()=>Promise<string> = ()=> {
+			return new Promise((resolve)=>{
+				return resolve(`/* custom template async */`)
+			})
+		}
```
mmkal, line 125: You'll get lint errors here, fix with `npm run lint -- --fix`

Also this type annotation is awkward, let's drop it:

```suggestion
		const methodAsync = async () => {
			return `/* custom template async */`
		}
```

And last minor thing, let's call it `template` rather than `methodAsync`, it would be good to keep the tests as realistic as possible.

mmkal (changes requested):

mmkal (approved):

PR #602, opened 2023-04-17
feat: remove uniqueness from sequelize storage

mau31415:
…tries in case of rerun

### DIFF
--- a/test/storage/sequelize.test.ts
+++ b/test/storage/sequelize.test.ts
@@ -66,20 +66,27 @@ describe('sequelize', () => {
 				.then(describeModel)
 				.then((description: any) => {
 					expect(description).toMatchInlineSnapshot(`
-						{
-						  "name": {
-						    "allowNull": false,
-						    "defaultValue": undefined,
-						    "primaryKey": true,
-						    "type": "VARCHAR(255)",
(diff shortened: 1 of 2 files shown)


danAlmogSB (approved):
useful change

mmkal:
Thanks. Could you add a test that fails without this change?

test/storage/sequelize.test.ts
```diff
@@ -266,6 +287,19 @@ describe('sequelize', () => {
...
+				sequelize: helper.sequelize,
+			});
+
+			await storage.logMigration({ name: 'm1' });
+			expect(await storage.executed()).toEqual(['m1']);
+
+			await storage.logMigration({ name: 'm1' });
+			await storage.logMigration({ name: 'm2' });
+			expect(await storage.executed()).toEqual(['m1', 'm1', 'm2']);
+		});
```
mmkal, line 302: Thank you for adding this test, it makes it clearer. Are we losing some safety by making it non-unique by default though?

jbalatero:
Since this was reverted, how do we work around --rerun ALLOW now?

mau31415:
I solved it on my end, I didn't have time to create a better fix. Sorry...

PR #613, opened 2023-08-02
feat: non-breaking ESM support 🤝 

### LINKED ISSUE #608: Migrate to ESM-first
Hey!

So this may seem like a feature request, but it actually falls under the bug category in my opinion.

As of today, `umzug` is written and built in CJS.
The problem is, this breaks applications written in pure ESM because `umzug` is using a dynamic `require`.
This problem is currently affecting several ORMs (including `sequelize`).

I think migrating to ESM-first is necessity since ESM supports importing any kind of module unlike CJS.

I am willing to write a PR to that effect if reviewers are available to check it.
This will obviously create many changes within this dependency. It should not, however, create any breaking change.
Let me know if you are willing to migrate and if you want a PR implementing this change.

mmkal:
Closes #608

This adds some esm-vs-commonjs "smartness" to umzug:

1. use `require` for `.cjs` migrations and `import` for `.mjs` migrations (and their typescript equivalents)
2. use `require` for `.js` migrations _if_ `typeof require.main === 'object'`, and `import` to `.js` migrations otherwise
3. use the same criteria to create (c)js vs mjs templates when creating migration files
4. add `"moduleResolution": "Node16"` to tsconfig.lib.json to make sure `import(filepath)` doesn't get transpiled into `__importStar(require(filepath))` (see [here](https://github.com/microsoft/TypeScript/issues/43329#issuecomment-926931937) and [here](https://github.com/microsoft/TypeScript/issues/43329#issuecomment-1276000768))

Tests/examples:

- add a `vanilla-esm` example to make sure using `import` / top-level await works
- add a step to the `test_pkg` job to make sure vitest isn't hiding gnarly import problems - this is installing the compiled library as a `.tgz`, and with no other dev/prod dependencies like vitest or ts-node having been installed, so should be very close to what end users will do

Didn't:

- add a wrapper.mjs file in the compiled folder as suggested in https://github.com/sequelize/umzug/issues/608#issuecomment-1611845497, mostly just because it didn't seem to be necessary? It seems to work fine when imported from an ES-module, using top-level await, etc., even though umzug is itself a commonjs module.

<details>
<summary>original body</summary>

~Related to #608 - although does not close it.~

~This adds built-in support for `.mjs` and `.mts` files. `.mjs` should "just work" - write migrations as ESM modules and they'll be imported in the right way. For the current major version, `.js` will continue to be assumed commonjs. ESM-fans will just have to type that extra `m` in their filenames.~

~This PR _doesn't_ add a wrapper file so that the umzug library itself can be imported as an ES module. That can be done in a follow-up PR. In the meantime, ESM users can use `createRequire` as in the [existing ESM example](https://github.com/sequelize/umzug/tree/main/examples/2.es-modules).~
</details>


mmkal:
Converted to draft - there is more work needed. When typescript compiles for commonjs, it replaces even dynamic imports with `require(...)`. So this works in tests but won't work when built. So, will either need to compile completely differently, or to use `eval('import')(filepath)` or something. I'll see if it's possible to avoid `eval` without doing a major rewrite.

mmkal:
@ephys @SmashingQuasar @JSteunou @mysuf since you've all commented on #608 and/or #626. I've published this PR, which I hope adds what we can consider "first-class" ESM support, as [3.5.0-0](https://www.npmjs.com/package/umzug/v/3.5.0-0) (under the `beta` tag). Having tested it, it seems to work as expected even without a wrapper `.mjs` file in the `lib/` directory.

If you're able, would you be able to try this version and see if it works for you? It does not migrate to ESM-first, but it now uses `import(...)` instead of `require(...)` when it detects ESM migrations. I also added some tests to make sure `import { Umzug } from 'umzug'` and top-level await work, etc.

Worth reading in the PR description, but this doesn't change umzug to be a `"type": "module"` package itself. IMO, there's no need for that, even in 2023. As long as ESM users can use it without having to write custom resolvers etc.

SmashingQuasar:
> @ephys @SmashingQuasar @JSteunou @mysuf @papb since you've all commented on #608 and/or #626. I've published this PR, which I hope adds what we can consider "first-class" ESM support, as [3.5.0-0](https://www.npmjs.com/package/umzug/v/3.5.0-0) (under the `beta` tag). Having tested it, it seems to work as expected even without a wrapper `.mjs` file in the `lib/` directory.
> 
> If you're able, would you be able to try this version and see if it works for you? It does not migrate to ESM-first, but it now uses `import(...)` instead of `require(...)` when it detects ESM migrations. I also added some tests to make sure `import { Umzug } from 'umzug'` and top-level await work, etc.
> 
> Worth reading in the PR description, but this doesn't change umzug to be a `"type": "module"` package itself. IMO, there's no need for that, even in 2023. As long as ESM users can use it without having to write custom resolvers etc.

You efforts are very much appreciated. Thanks a lot for taking time to look into this. It will benefit everyone in the long run!
I'll have a look whenever possible, we are a bit behind right now so it may not be an immediate thing.

mysuf:
@mmkal Thank you. I've tried and in my case it works out of the box without custom resolver as expected.

zevisert:
`3.5.0-0` worked in my project that's working on an ESM migration with no other changes as well!

JSteunou:
@mmkal great job, really appreciate the effort! :clap:

Issue #623, opened 2023-10-26
Do not set charset and collate options on postgresql


QuentinFarizon:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `umzug@2.3.0` for the project I'm working on.

Sequelize v7.0.0-alpha.33 complains (rightly so), that collate and charset optinos are not supposed to be passed for postgres dialect. To my knowledge, this is only used for mysql

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/umzug/lib/storages/SequelizeStorage.js b/node_modules/umzug/lib/storages/SequelizeStorage.js
index 1277063..469a6cc 100644
--- a/node_modules/umzug/lib/storages/SequelizeStorage.js
+++ b/node_modules/umzug/lib/storages/SequelizeStorage.js
@@ -90,8 +90,8 @@ class SequelizeStorage extends _Storage.default {
       tableName: this.tableName,
       schema: this.schema,
       timestamps: this.timestamps,
-      charset: 'utf8',
-      collate: 'utf8_unicode_ci'
+      charset: this.sequelize.dialect.name === 'mysql' ? 'utf8' : null,
+      collate: this.sequelize.dialect.name === 'mysql' ? 'utf8_unicode_ci' : null
     });
   }
   /**
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

WikiRik:
See #612 for earlier discussion about this

QuentinFarizon:
@WikiRik What do you think of my solution ? 

With something like :
```
const charsetableDialects = ['mysql', 'mariadb', 'snowflake']
(...)
charset: charsetableDialects.includes(this.sequelize.dialect.name) ? 'utf8' : null,
collate: charsetableDialects.includes(this.sequelize.dialect.name) ? 'utf8_unicode_ci' : null
```

WikiRik:
That looks good

QuentinFarizon:
@WikiRik I've pushed a PR : https://github.com/sequelize/umzug/pull/624

qfarizon-qftech:
@WikiRik It would be great to update sequelize-cli to use this new version !

WikiRik:
[...]

Issue #612, opened 2023-07-25
SequelizeStorage model uses options that are rejected by sequelize v7 + postgres et al


chriswheeldon-peakon:
The SequelizeStorage engine is specifying `charset` and `collate` in the table definition options [here](https://github.com/sequelize/umzug/blob/main/src/storage/sequelize.ts#L145). I think these options are meaningless to postgres among others and, since https://github.com/sequelize/sequelize/pull/15846, are now rejected by sequelize with an exception when umzug attempts to synchronise its metadata table.

Versions:
umzug: 3.2.1
Storage: Sequelize
@sequelize/core: 7.0.0-alpha.27 
postgres: 15.2

n.b. it is possible to work around this by passing an already created sequelize model definition to the SequelizeStorage constructor e.g.

```js
new SequelizeStorage({
    model: sequelize.isDefined(tableName)
      ? sequelize.model(tableName)
      : sequelize.define(
          tableName,
          {
            [columnName]: {
              type: DataTypes.STRING,
              allowNull: false,
              unique: true,
              primaryKey: true,
              autoIncrement: false,
            },
          },
          {
            tableName,
            timestamps: false,
            schema,
          }
        ),
  });
```

chriswheeldon-peakon:
As far as I can tell, from the sequelize PR, only the `snowflake` and, through their absence in the PR,  `mariadb` and `mysql` dialects support these options.

mmkal:
Thank you! Maybe we can allow options overrides so v7 users can do something like 

```ts
new SequelizeStorage(..., {
  options: ({charset, collate, ...options}) => options,
})
```

or something like that. 

Then, at some point, v7 could become the default, or another storage could be added for it or something.

EinfachHans:
With latest v7 alpha i get this error:

`Error: The following options are not supported by createTableQuery in postgres: charset, collate`

Setting the values to `undefined` instead of `null` works for me - Should i create a pr?

PR #624, opened 2023-10-27
Provide charset and collate options only for dialects supporting them

QuentinFarizon:
Fixes https://github.com/sequelize/umzug/issues/612 

Fixes https://github.com/sequelize/umzug/issues/623

### DIFF
--- a/src/storage/sequelize.ts
+++ b/src/storage/sequelize.ts
@@ -27,7 +27,12 @@ export type SequelizeType = {
 	isDefined(modelName: string): boolean;
 	model(modelName: string): any;
 	define(modelName: string, columns: {}, options: {}): {};
+	dialect?: {
+		name?: string;
+	};
 };
+
+const DIALECTS_WITH_CHARSET_AND_COLLATE = new Set(['mysql', 'mariadb', 'snowflake']);
 
 type ModelClassType = ModelClass & (new (values?: object, options?: any) => ModelTempInterface);
 
@@ -127,6 +132,9 @@ export class SequelizeStorage implements UmzugStorage {
 			return this.sequelize.model(this.modelName);
 		}
 
+		const dialectName = this.sequelize.dialect?.name;
+		const hasCharsetAndCollate = dialectName && DIALECTS_WITH_CHARSET_AND_COLLATE.has(dialectName);
+
 		return this.sequelize.define(
 			this.modelName,
 			{
@@ -142,8 +150,8 @@ export class SequelizeStorage implements UmzugStorage {
 				tableName: this.tableName,
 				schema: this.schema,
 				timestamps: this.timestamps,
-				charset: 'utf8',
-				collate: 'utf8_unicode_ci',
+				charset: hasCharsetAndCollate ? 'utf8' : null,
+				collate: hasCharsetAndCollate ? 'utf8_unicode_ci' : null,
 			}
 		) as ModelClassType;
 	}


WikiRik:
You'll need to update the minimal type here to add the dialect property;  https://github.com/sequelize/umzug/blob/main/src/storage/sequelize.ts#L21-L30

src/storage/sequelize.ts
```diff
@@ -16,6 +16,9 @@ export type ModelClass = {
 	findAll(options?: {}): Promise<any[]>;
 	create(options: {}): Promise<void>;
 	destroy(options: {}): Promise<void>;
+	dialect: {
+		name: string;
+	};
```
WikiRik, line 21: You'll need to add this to SequelizeType I think. Did that give different errors?

QuentinFarizon: Sorry, fixed the typo

src/storage/sequelize.ts
```diff
@@ -27,8 +27,13 @@ export type SequelizeType = {
 	isDefined(modelName: string): boolean;
 	model(modelName: string): any;
 	define(modelName: string, columns: {}, options: {}): {};
+	dialect: {
+		name: string;
```
WikiRik, line 31:
```suggestion
	dialect?: {
		name?: string;
```

Make these optional so the other tests won't complain

src/storage/sequelize.ts
```diff
@@ -142,8 +147,8 @@ export class SequelizeStorage implements UmzugStorage {
 				tableName: this.tableName,
 				schema: this.schema,
 				timestamps: this.timestamps,
-				charset: 'utf8',
-				collate: 'utf8_unicode_ci',
+				charset: DIALECTS_WITH_CHARSET_AND_COLLATE.includes(this.sequelize.dialect.name) ? 'utf8' : null,
+				collate: DIALECTS_WITH_CHARSET_AND_COLLATE.includes(this.sequelize.dialect.name) ? 'utf8_unicode_ci' : null,
```
WikiRik, line 151:
```suggestion
				charset: DIALECTS_WITH_CHARSET_AND_COLLATE.includes(this.sequelize.dialect!.name!) ? 'utf8' : null,
				collate: DIALECTS_WITH_CHARSET_AND_COLLATE.includes(this.sequelize.dialect!.name!) ? 'utf8_unicode_ci' : null,
```

These are always given in the actual sequelize code, just not in our simple mock type

WikiRik:
We're one step further, now the linting issues commence

src/storage/sequelize.ts
```diff
@@ -142,8 +147,8 @@ export class SequelizeStorage implements UmzugStorage {
 				tableName: this.tableName,
 				schema: this.schema,
 				timestamps: this.timestamps,
-				charset: 'utf8',
-				collate: 'utf8_unicode_ci',
+				charset: DIALECTS_WITH_CHARSET_AND_COLLATE.has(this.sequelize.dialect!.name!) ? 'utf8' : null,
+				collate: DIALECTS_WITH_CHARSET_AND_COLLATE.has(this.sequelize.dialect!.name!) ? 'utf8_unicode_ci' : null,
```
mmkal, line 151: I agree with the the linter here. If dialect is an optional property we shouldn't use !.

Let's make it a variable and null check it:

```ts
const dialectName = this.sequelize.dialect?.name
const hasCharsetAndCollate = dialectName && DIALECTS_WITH_CHARSET_AND_COLLATE.has(dialectName)

// ...

charset: hasCharsetAndCollate ? 'utf8' : null,
collate: hasCharsetAndCollate ? 'utf8_unicode_ci' : null,
```

mmkal (changes requested):

mmkal (approved):

PR #641, opened 2024-01-10
Snowflake does not explicitly support collation in the same way as some other databases

lbazetto:
Snowflake does not explicitly support collation in the same way as some other databases.

Like it shows in [create table syntax](https://docs.snowflake.com/en/sql-reference/sql/create-table#syntax)
And in [collation specification](https://docs.snowflake.com/en/sql-reference/collation#label-collation-specification)

So this fix removes Snowflake from `DIALECTS_WITH_CHARSET_AND_COLLATE`

### DIFF
--- a/src/storage/sequelize.ts
+++ b/src/storage/sequelize.ts
@@ -32,7 +32,7 @@ export type SequelizeType = {
 	};
 };
 
-const DIALECTS_WITH_CHARSET_AND_COLLATE = new Set(['mysql', 'mariadb', 'snowflake']);
+const DIALECTS_WITH_CHARSET_AND_COLLATE = new Set(['mysql', 'mariadb']);
 
 type ModelClassType = ModelClass & (new (values?: object, options?: any) => ModelTempInterface);
 


lbazetto:
> In sequelize itself there is support for charset and collation when creating a database and table. Is that also incorrect? If so; could you make a corresponding PR for that?
> https://github.com/sequelize/sequelize/blob/962e5c05c0ac92c7e6b5140ae56a3e12ba8165e7/packages/core/src/dialects/snowflake/query-generator-typescript.ts#L19
> https://github.com/sequelize/sequelize/blob/962e5c05c0ac92c7e6b5140ae56a3e12ba8165e7/packages/core/src/dialects/snowflake/query-generator.js#L32

Sure, PR has been created:
https://github.com/sequelize/sequelize/pull/16962

mmkal:
@QuentinFarizon since you added this list, does this look ok to you?

WikiRik (approved):
Support for charset and collate was indeed incorrectly added for snowflake in sequelize itself and will be removed in a next v7 alpha. That should also be reflected here

QuentinFarizon:
@mmkal sure, I wasn't the one suggesting adding it

PR #644, opened 2024-01-22
Fix DeprecationWarning for Sequelize V7

anroypaul:
This PR is fixing DeprecationWarning for Sequelize V7 (currently alpha, warning [SEQUELIZE0010]): Accessing DataTypes on the Sequelize constructor is deprecated. Use the DataTypes object instead. 

Reference to [Sequelize V7 upgrade](https://sequelize.org/docs/v7/other-topics/upgrade/#new-deprecations).


WikiRik (approved):
DataTypes.STRING does also exist for v6 so this is not a breaking change or anything

PR #658, opened 2024-03-28
bump glob to v10

### LINKED ISSUE #638: Snyk find issue in umzug@3.4.0 > glob@8.1.0 > inflight@1.0.6
```sh
$ snyk test
Issues with no direct upgrade or patch:
  ✗ Missing Release of Resource after Effective Lifetime [Medium Severity][https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116] in inflight@1.0.6
    introduced by umzug@3.4.0 > glob@8.1.0 > inflight@1.0.6
  No upgrade or patch available
```

mmkal:
closes #638

sort paths since they come back in reverse order now

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
 	"dependencies": {
 		"@rushstack/ts-command-line": "^4.12.2",
 		"emittery": "^0.13.0",
-		"glob": "^8.0.3",
+		"glob": "^10.3.10",
 		"pony-cause": "^2.1.4",
 		"type-fest": "^4.0.0"
 	},
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,8 +12,8 @@ dependencies:
     specifier: ^0.13.0
     version: 0.13.1
   glob:
-    specifier: ^8.0.3
-    version: 8.1.0
+    specifier: ^10.3.10
+    version: 10.3.10
   pony-cause:
     specifier: ^2.1.4
     version: 2.1.10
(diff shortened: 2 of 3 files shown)


WikiRik:
@mmkal Glob v10 does not support Node 12 as per the following;
https://github.com/isaacs/node-glob/blob/b5d0f640893eba729bb8675a51a73ffbc0760b35/package.json#L96

Are you sure the tests on Node 12 passed?

mmkal:
It looks like there's a bug that's making the Node 12 and Node 14 `npm test` commands pass even though they're both failing:

image

I will look into that, and maybe just switch to [fast-glob](https://www.npmjs.com/package/fast-glob?activeTab=code) which supports nodejs 8+.

B4nan:
Why would you care about this? Ship a major version if the current one is supposed to work with this. Supporting node 16 is questionable (but understandable), supporting anything older is (IMO) just weird and a waste of (your/maintainer's) time. 

Node 12 is not receiving security updates for 2 years now, node 14 for a year, and node 16 for more than 6 months.

WikiRik:
> Why would you care about this? Ship a major version if the current one is supposed to work with this. 

If we are shipping this in a major version then it's not an issue and we can use glob v10 just fine. But as far as I know this was supposed to be released as a minor or patch version, which would break older projects and not be in the spirit of semantic versioning.

mmkal:
I prefer not to release major versions unless there's a good reason. Especially for a library like this which is sometimes a couple layers deep in dependency trees. Using fast-glob is an easy fix, and not worth a major version IMO. The tests for lower node versions were just broken.

Having said that, eventually it will be too much of a headache to support old node versions. At that point I'm fine with a major bump.

PR #660, opened 2024-03-28
fix: restore node12 support (glob -> fast-glob)

### LINKED ISSUE #659: Glob v10 does not support Node 12
_Originally posted by @WikiRik in https://github.com/sequelize/umzug/issues/658#issuecomment-2024450009_

Suggested fix: switch to fast-glob

`glob` has a helpful [comparison section](https://github.com/isaacs/node-glob?tab=readme-ov-file#comparison-to-other-javascript-glob-implementations):

<blockquote>
# Comparison to Other JavaScript Glob Implementations

If you want glob matching that is as faithful as possible to Bash pattern expansion semantics, and as fast as possible within that constraint, use this module.
If you are reasonably sure that the patterns you will encounter are relatively simple, and want the absolutely fastest glob matcher out there, use [fast-glob](http://npm.im/fast-glob).
If you are reasonably sure that the patterns you will encounter are relatively simple, and want the convenience of automatically respecting .gitignore files, use [globby](http://npm.im/globby).
There are some other glob matcher libraries on npm, but these three are (in my opinion, as of 2023) the best.
</blockquote>

I think it's pretty safe to assume the migration globs are simple enough for fast-glob to work for the vast majority of users. If not, they're hopefully advanced users and could always pass an array of migrations built via `glob` anyway.

mmkal:
fixes #659

Note that there are some edge-case differences between glob and fast-glob, but it's very unlikely that users of umzug will hit them.

See glob docs for an in-depth comparison: https://github.com/isaacs/node-glob?tab=readme-ov-file#comparison-to-other-javascript-glob-implementations

If there's anyone that needs the old `glob` behaviour, they can use glob themselves and pass an array of migrations to umzug. (see https://github.com/sequelize/umzug#direct-migrations-list)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
           node migrate create --name new-migration.js
           node migrate up
       - name: run vanilla esm example
+        if: matrix.node != 12
         working-directory: examples/0.5-vanilla-esm
         run: |
           npm init -y
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ const umzug = new Umzug({
 
 Note on migration file sorting:
 
-- file matches, found using [glob](https://npmjs.com/package/glob), will be lexicographically sorted based on their paths
+- file matches, found using [fast-glob](https://npmjs.com/package/fast-glob), will be lexicographically sorted based on their paths
   - so if your migrations are `one/m1.js`, `two/m2.js`, `three/m3.js`, the resultant order will be `one/m1.js`, `three/m3.js`, `two/m2.js`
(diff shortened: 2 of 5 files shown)


mmkal:
CC @WikiRik any feedback on this change?

WikiRik:
I don't have enough insights in the usages of umzug and the globs people use. So I can't say much on this.

It seems that there are well documented changes between the two libraries that people can refer to and the added benefit is that the indirect vulnerable dependency is resolved without dropping Node 12 support.

mmkal:
Agreed. I am working on a change to CI which will highlight that node 12/node 14 support is broken currently. I'll merge that first, which should make CI go red, then confirm that this change fixes it.

PR #694, opened 2024-10-28
fix: logMigration breaking on rerun ALLOW for mssql

semics-tech:
When running the  `up`  command with `--name` and `--rerun ALLOW` arguments, the `logMigration` function is failing due to it trying to insert a record that already exists, resulting in a `ValidationError`.

This code checks for this and don't re-run the logMigration but you may want to update or do something else instead?


semics-tech:
@mmkal I see you were the last person to complete a PR, would you kindly be able to review approve this please? It's blocking me from being able to re-apply migrations.

PR #697, opened 2025-02-01
Fix "can not convert object to primitive" nonsence error message

secam:
MikroOrm that uses umzug under the hood, throws custom errors in the form of objects derived from null (Null prototype objects) that are not handled correctly by errorString method, this result in confusing error message masking original error completely (nothing related to original error in stack trace)
This micro fix solves this and makes errorString more universal in handling such tricky objects.

### DIFF
--- a/src/umzug.ts
+++ b/src/umzug.ts
@@ -48,9 +48,14 @@ export class MigrationError extends errorCause.ErrorWithCause<unknown> {
   }
 
   private static errorString(cause: unknown) {
-    return cause instanceof Error
-      ? `Original error: ${cause.message}`
-      : `Non-error value thrown. See info for full props: ${cause as string}`
+    if (cause instanceof Error) return `Original error: ${cause.message}`
+    const msg = 'Non-error value thrown. See info for full props'
+    try {
+      return `${msg}: ${String(cause)}`
+    } catch {
+      //Null prototype object 'cause' would end up here
+      return msg
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


test/umzug.test.ts
```diff
@@ -705,7 +708,7 @@ describe('alternate migration inputs', () => {
     )
 
     await expect(umzug.down()).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Migration m1 (down) failed: Non-error value thrown. See info for full props: Some cryptic failure"`,
```
mmkal, line 708: This removed test case is still important. Instead of changing it could you copy paste the test and update the duplicated one.

src/umzug.ts
```diff
@@ -48,9 +48,14 @@ export class MigrationError extends errorCause.ErrorWithCause<unknown> {
...
-      ? `Original error: ${cause.message}`
-      : `Non-error value thrown. See info for full props: ${cause as string}`
+    if (cause instanceof Error) return `Original error: ${cause.message}`
+    const msg = 'Non-error value thrown. See info for full props'
+    try {
+      return `${msg}: ${String(cause)}`
+    } catch {
+      //Null prototype object 'cause' would end up here
+      return msg
+    }
```
mmkal, line 58: Is there a way to check explicitly for null prototype object without a try/catch?

test/umzug.test.ts
secam: Sure, restored the original test case and duplicated for null prototype with correct description and both up/down

src/umzug.ts
secam: Yep, did that, it involves some sort of "unfair" object casting for TS but seems to be working fine. Point for try catch was to be more bulletproof, but I agree it comes with performance penalty.

